PR #6, opened 2024-05-14
Client Refactoring

Mo-Fatah:
Main changes:
- Created `api_handlers.go` (open to any better names) in the client package to include all gateways to interact with Yunikorn API. Before that, all calls to Yunikorn APIs were within the `load**` functions. I removed them and make the gateway more formal because we will need these API calls later if we agreed to follow the pattern mentioned here https://github.com/G-Research/gr-oss/issues/654#issuecomment-2110262779


internal/repository/application.go
```diff
@@ -18,19 +18,19 @@ func (s *RepoPostgres) UpsertApplications(ctx context.Context, apps []*dao.Appli
...
-			requests = EXCLUDED.requests,
-			allocations = EXCLUDED.allocations,
-			state = EXCLUDED.state,
-			rejected_message = EXCLUDED.rejected_message,
-			state_log = EXCLUDED.state_log,
-			place_holder_data = EXCLUDED.place_holder_data,
-			has_reserved = EXCLUDED.has_reserved,
-			reservations = EXCLUDED.reservations,
-			max_request_priority = EXCLUDED.max_request_priority`
+			used_resource = COALESCE(EXCLUDED.used_resource, applications.used_resource),
```
Mo-Fatah, line 21: those might be generalized to all upserts in an upcoming PRs

internal/repository/postgres.go
```diff
@@ -37,11 +37,11 @@ func (s *RepoPostgres) Setup(ctx context.Context) {
 			id UUID,
 			cluster_id TEXT NOT NULL,
 			name TEXT NOT NULL,
-			capacity JSONB NOT NULL,
-			used_capacity JSONB NOT NULL,
-			utilization JSONB NOT NULL,
```
Mo-Fatah, line 42: I don't find a need to be strict here, those can cause unnecessary errors

internal/ykclient/client.go
```diff
@@ -36,9 +33,18 @@ var (
 	// nodeEndPt  = func(partitionName, nodeID string) string {
 	// 	return fmt.Sprintf("/ws/v1/partition/%s/node/%s", partitionName, nodeID)
 	// }
-	nodeUtilEndPt     = "/ws/v1/scheduler/node-utilizations"
-	applicationsEndPt = func(partitionName, queueName string) string {
-		return fmt.Sprintf("/ws/v1/partition/%s/queue/%s.%s/applications", partitionName, queueName, partitionName)
```
Mo-Fatah, line 41: I checked with creating multiple queues, it seems that the partition name isn't included in the queue path. Usually when we test our setup with a single queue, `root`, Yunikorn creates a default leaf queue called `root.default` because jobs can only be submitted to a leaf queue so it creates `root.default` as a default behaviour.

internal/ykclient/client.go
```diff
@@ -36,9 +33,18 @@ var (
...
+	nodeUtilEndPt           = "/ws/v1/scheduler/node-utilizations"
+	applicationsByPartEndPt = func(partitionName string) string {
+		return fmt.Sprintf("/ws/v1/partition/%s/applications/active", partitionName)
```
Mo-Fatah, line 38: for the `/active` part https://yunikorn.apache.org/docs/next/api/scheduler/#partition-applications

internal/ykclient/api_handlers.go
```diff
@@ -0,0 +1,296 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/apache/yunikorn-core/pkg/webservice/dao"
+)
+
+var (
+	streamEndPt            = "/ws/v1/events/stream"
```
richscott, line 15: Maybe eventually we should move these into shared definitions in some place like a `internal/config/constants.go` or some such? The `yunikorn-k8shim` repo puts lots of endpoints in its `test/e2e/framework/configmanager/constants.go`, for example.

richscott (approved):

PR #16, opened 2024-05-21
Accept base directory of ynikorn repos as cmd argument.

sudiptob2:
It seems convenient to take the `base_dir` as cmd arg, rather than manually modifying the `yk-deploy.sh` if the base directory is different.


sudiptob2:
I will update the developer docs if this PR is accepted 🙂

richscott (approved):
A great improvement!

Mo-Fatah:
> I will update the developer docs if this PR is accepted 🙂
@sudiptob2 that will be great! thanks!

PR #37, opened 2024-06-07
Add pgadmin container in docker-compose.

sudiptob2:
A pgadmin container will allow quick and convenient lookup to the development database 😊


developer/docker-compose.yml
```diff
@@ -3,8 +3,6 @@ services:
     container_name: postgres
     image: postgres:15.2
```
richscott, line 4: As long as we're updating this file, can you update this to 15.7 ? (I think that's the latest minor release in Postgres 15).

richscott (commented):
Just one suggestion/request.

developer/docker-compose.yml
sudiptob2: sure

dave-gantenbein (approved):

Issue #46, opened 2024-06-13
Make configuration parameter definiton consistent.


sudiptob2:
In the current state of `yhs`, we can declare environment variables in `yhs.yaml` or as ENV vars. However, some configuration parameters defined in `yhs.yaml` cannot be consistently defined as ENV vars. This issue is related to how we parse `yhs.yaml` and ENV vars using `koanf`. 

For example, consider the following configuration:
https://github.com/G-Research/yunikorn-history-server/blob/bbd4467a2bb16fbeca1e745df57d9a963966e36e/yhs.yml#L16-L17
With the current logic, we cannot define an ENV var that will be parsed as `yhs.serverAddr` by `koanf`.

**Example of how ENVs are parsed currently**

`YHS_PARENT1_CHILD1_NAME` will be converted into `parent1.child1.name`

**Solution 1:**  
Modify the logic for parsing environment variables to accommodate various combinations of uppercase, lowercase, and snake_case formats. Ref: [loadConfig()](https://github.com/G-Research/yunikorn-history-server/blob/bbd4467a2bb16fbeca1e745df57d9a963966e36e/cmd/event-collector/main.go#L31)

**Solution 2:**  
Change the way configuration parameters are defined in the `yhs.yaml` file. Ensure consistency when selecting new variable names, and consider how the respective environment variables should be formatted.

**Opinion:**  
I believe Solution 2 is better because environment variables are typically defined using uppercase letters with underscores. Therefore, we should update the `yhs.yaml` file accordingly.

PR #50, opened 2024-06-13
Consistently process env vars and config parameter from yhs

### LINKED ISSUE #45: bug 🐛: environment variable used incorrectly.
The environment variables in the following code are used incorrectly. In `yhs.yaml` they are defined under `db`. So should be parsed as `db.env_name`.
https://github.com/G-Research/yunikorn-history-server/blob/bbd4467a2bb16fbeca1e745df57d9a963966e36e/cmd/event-collector/main.go#L81-L92

sudiptob2:
Closes #45 #46

Changes:

- [x] Update logic of `loadConfig()`
- [x] Make all configuration parameter definitions consistent
- [x] Bugfix: improper read of env parameters #45
- [x] Update unit tests

### DIFF
--- a/cmd/event-collector/main.go
+++ b/cmd/event-collector/main.go
@@ -28,6 +28,17 @@ var (
 	eventCounts   config.EventTypeCounts
 )
 
+// Removes the prefix "YHS_" and replaces the first "_" with "."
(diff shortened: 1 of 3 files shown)


richscott (approved):

sudiptob2:
Thanks @richscott 🚀

PR #52, opened 2024-06-13
Configuration support for zap logger

### LINKED ISSUE #48: Add support for logging settings to config file/env vars
The code that adds logging support to YHS, introduced in https://github.com/G-Research/yunikorn-history-server/pull/28/files, has several constant variables for various logging settings. These should be read from the YHS configuration file (typically, `yhs.yml`) or env vars, like other YHS parameters. Currently, the logging settings are:
```
logFilePath = "yhs.log"
maxSize     = 5
maxBackups  = 10
maxAge      = 14
compress    = true
logLevel    = zap.InfoLevel
```
These should be put into a block, perhaps named `logging`.  Also, the log level value should probably be specifiable in the format of `INFO`, `WARN`, etc, rather than referencing the zap package level.

sudiptob2:
closes #48 

Changes:

- [x] Reading logging configuration for ENV/config file
- [x] Support for Log level strings `DEBUG, INFO, WARN, ERROR, DPANIC, PANIC, FATAL`
- [x] Some housekeeping

### DIFF
--- a/charts/values.yaml
+++ b/charts/values.yaml
@@ -56,15 +56,7 @@ yunikorn:
   protocol: "http"
 
 log:
-    # -- Log file path
-    filePath: "yhs.log"
-    # -- Maximum log file size in megabytes
-    maxSize: 5
-    # -- Maximum number of log file backups
-    maxBackups: 10
-    # -- Maximum age of log files in days
-    maxAge: 30
-    # -- Compress log files
-    compress: true
+    # -- Output type of the log, if true, log will be output in json format
+    production: true
     # -- Log level, one of DEBUG, INFO, WARN, ERROR, DPANIC, PANIC, FATAL
     level: "INFO"
--- a/charts/values.yaml
+++ b/charts/values.yaml
@@ -57,6 +57,6 @@ yunikorn:
 
 log:
     # -- Output type of the log, if true, log will be output in json format
-    production: true
+    useJSONFormat: true
     # -- Log level, one of DEBUG, INFO, WARN, ERROR, DPANIC, PANIC, FATAL
     level: "INFO"
--- a/charts/values.yaml
+++ b/charts/values.yaml
@@ -57,6 +57,6 @@ yunikorn:
 
 log:
     # -- Output type of the log, if true, log will be output in json format
-    useJSONFormat: true
+    jsonFormat: true
     # -- Log level, one of DEBUG, INFO, WARN, ERROR, DPANIC, PANIC, FATAL
     level: "INFO"
--- a/charts/values.yaml
+++ b/charts/values.yaml
@@ -54,3 +54,17 @@ yunikorn:
   port: ""
   # -- Yunikorn scheduler protocol
   protocol: "http"
+
+log:
+    # -- Log file path
+    filePath: "yhs.log"
+    # -- Maximum log file size in megabytes
+    maxSize: 5
+    # -- Maximum number of log file backups
+    maxBackups: 10
+    # -- Maximum age of log files in days
+    maxAge: 30
+    # -- Compress log files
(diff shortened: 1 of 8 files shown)


richscott (approved):

dejanzele:
Pods in Kubernetes are ephemeral, and we don't use any persistent volumes to backup log files. In case of a Pod restart, the log files will be lost.

Best practice is to log directly to STDOUT and STDERR and leave log managament, rotation... to Kubernetes.
Kubernetes in any case saves the logs from STD in a file.

No reason to keep a log file.

dejanzele:
It would be nice to add a config option to switch between human-readable logs and JSON logs.
In Kubernetes, JSON logs should be default.

log/logger.go
```diff
@@ -2,38 +2,37 @@ package log
...
+func InitLogger(config LogConfig) {
 	once.Do(func() {
 		stdout := zapcore.AddSync(os.Stdout)
 		file := zapcore.AddSync(&lumberjack.Logger{
-			Filename:   logFilePath,
-			MaxSize:    maxSize,
-			MaxBackups: maxBackups,
-			MaxAge:     maxAge,
-			Compress:   compress,
+			Filename:   config.LogFilePath,
```
dejanzele, line 31: check my comment, this is redundant

sudiptob2: Updated

log/logger.go
```diff
@@ -42,13 +30,14 @@ func InitLogger(config LogConfig) {
 		developmentCfg := zap.NewDevelopmentEncoderConfig()
 		developmentCfg.EncodeLevel = zapcore.CapitalColorLevelEncoder
 
-		consoleEncoder := zapcore.NewConsoleEncoder(developmentCfg)
-		fileEncoder := zapcore.NewJSONEncoder(productionCfg)
+		var encoder zapcore.Encoder
+		if config.IsProduction {
```
dejanzele, line 34: this can be simplified:
```
encoder = zapcore.NewConsoleEncoder(developmentCfg)
if config.IsProduction {
  encoder = zapcore.NewJSONEncoder(productionCfg)
}
```

dejanzele: I'd rename `IsProduction` to `UseJSONLogFormat`

sudiptob2: In that case, can we entirely remove the `development` config? And just keep only one configuration. And then according to the log format flag, we can switch the `Encoder`
```go
config := zap.NewProductionEncoderConfig()
var encoder zapcore.Encoder
encoder = zapcore.NewConsoleEncoder(config)
if config.IsProduction {
	encoder = zapcore.NewJSONEncoder(config)
}
```

dejanzele: that makes sense, I always try to avoid naming something in an application based on environment name and keep things generic

charts/values.yaml
```diff
@@ -53,4 +53,10 @@ yunikorn:
   # -- Yunikorn scheduler port
   port: ""
   # -- Yunikorn scheduler protocol
-  protocol: "http"
\ No newline at end of file
+  protocol: "http"
+
+log:
+    # -- Output type of the log, if true, log will be output in json format
+    useJSONFormat: true
```
dejanzele, line 60: I guess we could even shorten it to `jsonFormat`

log/logger.go
```diff
@@ -10,8 +10,8 @@ import (
 )
 
 type LogConfig struct {
-	UseJSONFormat bool
-	LogLevel      string
+	JsonFormat bool
```
dejanzele, line 13:
```suggestion
	JSONFormat bool
```

sudiptob2: updated

dejanzele (approved):

PR #60, opened 2024-06-15
Implement helm chart for YHS

### LINKED ISSUE #19: YHS: write helm chart(s) for installing/running Yunikorn History Server
Write one (or more if necessary, for different configurations) helm charts for installing and running YHS in a Kubernetes cluster. See https://helm.sh/ for more information.

sudiptob2:
closes #19

Implemented initial helm chart.


dejanzele:
Can you rename the folder `chart` -> `charts`?

chart/templates/configmap.yaml
```diff
@@ -0,0 +1,37 @@
...
+kind: ConfigMap
+metadata:
+  name: yhs-config
+  namespace: yunikorn
+  labels:
+    app: yunikorn-history-server
+    {{- with .Values.common.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+data:
```
dejanzele, line 25: Shouldn't this contain a `yhs.yaml` or `config.yaml` similarly how we use the config here https://github.com/G-Research/yunikorn-history-server/blob/main/yhs.yml

chart/templates/secret.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: yhs-secret
+  namespace: yunikorn
+  labels:
+    app: yunikorn-history-server
+    {{- with .Values.common.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+type: Opaque
+data:
+  YHS_DB_PASSWORD: "{{ required "A valid .secret.dbPassword is required!" .Values.secret.dbPassword | b64enc }}"
```
dejanzele, line 14: We should enable people to specify the password either inline (directly as a Helm parameter using `--set` which is insecure) or provide a `dbPasswordSecretRef` which would reference an already existing secret with this value.

chart/values.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+config:
+  # -- Yunikorn scheduler host
+  yunikornHost: ""
+  # -- Yunikorn scheduler port
+  yunikornPort: ""
+  # -- Yunikorn scheduler protocol
+  yunikornProtocol: "http"
+  # -- YHS database host
+  dbHost: ""
```
dejanzele, line 36: this can be groupd in a `db` block:
```
db:
  host:
  port:
```

chart/values.yaml
```diff
@@ -0,0 +1,59 @@
...
+  # -- Service type
+  type: LoadBalancer
+  # -- Service port
+  port: 8989
+  # -- Service target port
+  targetPort: 8989
+
+config:
+  # -- Yunikorn scheduler host
+  yunikornHost: ""
```
dejanzele, line 30: this can be grouped into a `yunicorn` block:
```
yunicorn:
  host:
```

chart/values.yaml
```diff
@@ -0,0 +1,59 @@
...
+  # -- Maximum number of connections in the database pool
+  dbPoolMaxConns: 0
+  # -- Minimum number of connections in the database pool
+  dbPoolMinConns: 0
+  # -- Maximum lifetime of a connection in the database pool
+  dbPoolMaxConnLifetime: 1800
+  # -- Maximum idle time of a connection in the database pool
+  dbPoolMaxConnIdleTime: 120
+  # -- YHS server address
+  yhsServerAddr: ":8989"
```
dejanzele, line 52: this is fine to be left as top level config

chart/values.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+
+secret:
```
dejanzele, line 55: I would remove this block, and in the `config` block provide support for defining `db.password` (inline) or `db.passwordSecretRef` which if provided will mount it from the provided secret

chart/values.yaml
sudiptob2: Yeah, initially I had done that but after reading [this](https://helm.sh/docs/chart_best_practices/values/#flat-or-nested-values) helm best practice guide I thought I should flatten as much as possible. I can group them again if grouping is better.

chart/templates/secret.yaml
sudiptob2: updated

chart/values.yaml
sudiptob2: Updated the structure of `values.yaml`

chart/values.yaml
sudiptob2: updated

chart/values.yaml
sudiptob2: `yhsServerAddr` moved to the top level

chart/values.yaml
sudiptob2: updated logic for db-pass

charts/values.yaml
```diff
@@ -0,0 +1,58 @@
...
+  # -- Docker image repository
+  repository: "gresearch/yunikorn-history-server"
+  # -- Docker image tag
+  tag: "latest"
+  # -- Docker image pull policy
+  pullPolicy: "IfNotPresent"
+
+service:
+  # -- Service type
+  type: LoadBalancer
```
dejanzele, line 24: Can you change the default to `ClusterIP`?

sudiptob2: updated

charts/values.yaml
```diff
@@ -0,0 +1,58 @@
...
+  # -- YHS database user
+  user: ""
+  # -- YHS database password (insecure, use secrets)
+  password: ""
+  # -- YHS database password secret reference
+  passwordSecretRef: ""
+  # -- Maximum number of connections in the database pool
+  poolMaxConns: 0
+  # -- Minimum number of connections in the database pool
+  oolMinConns: 0
```
dejanzele, line 46:
```suggestion
  poolMinConns: 0
```

charts/values.yaml
```diff
@@ -0,0 +1,58 @@
...
+  # -- YHS database name
+  name: ""
+  # -- YHS database user
+  user: ""
+  # -- YHS database password (insecure, use secrets)
+  password: ""
+  # -- YHS database password secret reference
+  passwordSecretRef: ""
+  # -- Maximum number of connections in the database pool
+  poolMaxConns: 0
```
dejanzele, line 44: what does 0 mean? unlimited?

charts/templates/configmap.yaml
```diff
@@ -0,0 +1,37 @@
+{{- $namespace := .Values.common.namespace | required "A valid .common.namespace is required!" -}}
```
dejanzele, line 1: This should be deleted, namespace is automatically overriden when users install it in a specific namespace with the `--namespace` flag

charts/templates/configmap.yaml
```diff
@@ -0,0 +1,37 @@
...
+{{- $dbPoolMinConns := .Values.db.poolMinConns | required "A valid .db.poolMinConns is required!" -}}
+{{- $dbPoolMaxConnLifetime := .Values.db.poolMaxConnLifetime | required "A valid .db.poolMaxConnLifetime is required!" -}}
+{{- $dbPoolMaxConnIdleTime := .Values.db.poolMaxConnIdleTime | required "A valid .db.poolMaxConnIdleTime is required!" -}}
+{{- $yhsServerAddr := .Values.yhsServerAddr | required "A valid .yhsServerAddr is required!" -}}
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: yhs-config
+  namespace: {{ $namespace }}
```
dejanzele, line 19: Omit namespace everywhere, it will be set with the `--namespace` flag during `helm install`

charts/templates/configmap.yaml
```diff
@@ -0,0 +1,37 @@
...
+{{- $dbPoolMaxConns := .Values.db.poolMaxConns | required "A valid .db.poolMaxConns is required!" -}}
+{{- $dbPoolMinConns := .Values.db.poolMinConns | required "A valid .db.poolMinConns is required!" -}}
+{{- $dbPoolMaxConnLifetime := .Values.db.poolMaxConnLifetime | required "A valid .db.poolMaxConnLifetime is required!" -}}
+{{- $dbPoolMaxConnIdleTime := .Values.db.poolMaxConnIdleTime | required "A valid .db.poolMaxConnIdleTime is required!" -}}
+{{- $yhsServerAddr := .Values.yhsServerAddr | required "A valid .yhsServerAddr is required!" -}}
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: yhs-config
```
dejanzele, line 18: Take a look at this `_helpers.tpl` file - https://github.com/armadaproject/armada-operator/blob/main/charts/armada-operator/templates/_helpers.tpl

Implement similar things for well-defined labels and name.

Take a look at this resource for example:
- https://github.com/armadaproject/armada-operator/blob/main/charts/armada-operator/templates/deployment.yaml#L10
- https://github.com/armadaproject/armada-operator/blob/main/charts/armada-operator/templates/webhook-service.yaml#L14
- https://github.com/armadaproject/armada-operator/blob/main/charts/armada-operator/templates/webhook-service.yaml#L4

charts/templates/deployment.yaml
```diff
@@ -0,0 +1,52 @@
...
+{{- $replicaCount := .Values.replicaCount | required "A valid .Values.replicaCount is required!" -}}
+{{- $imageRegistry := .Values.image.registry | required "A valid .Values.image.registry is required!" -}}
+{{- $imageRepository := .Values.image.repository | required "A valid .Values.image.repository is required!" -}}
+{{- $imageTag := .Values.image.tag | required "A valid .Values.image.tag is required!" -}}
+{{- $imagePullPolicy := .Values.image.pullPolicy | required "A valid .Values.image.pullPolicy is required!" -}}
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: "yunikorn-history-server"
```
dejanzele, line 16: similar things apply as the comment above

charts/templates/deployment.yaml
```diff
@@ -0,0 +1,52 @@
...
+  namespace: "yunikorn"
```
dejanzele, line 17: similar as above

charts/values.yaml
sudiptob2: Actually, I picked up these default values from the `yhs.yaml`

charts/values.yaml
sudiptob2: fixed

charts/templates/configmap.yaml
sudiptob2: removed explicit namespace definition in the chart.

charts/templates/configmap.yaml
sudiptob2: Implemented similar `_helpers.tpl` to manage the labels

dejanzele (approved):

dejanzele:
LGTM

Issue #15, opened 2024-05-20
Unit testing for `internal/ykclient/api_handlers.go`


Mo-Fatah:
`api_handlers.go` seems solid for now. 
Add some https://github.com/G-Research/yunikorn-history-server/blob/main/internal/ykclient/api_handlers.go that will handle some corner cases from the YK scheduler such as connection failures and `404`

richscott:
Fixed by https://github.com/G-Research/yunikorn-history-server/pull/64

PR #64, opened 2024-06-23
Test/15/test api handlers

sudiptob2:
related to #15

### DIFF
--- a/internal/yunikorn/rest_test.go
+++ b/internal/yunikorn/rest_test.go
@@ -127,20 +127,7 @@ func TestRESTClient_GetApplication(t *testing.T) {
 	}
 }
 
-func getMockServerYunikornConfig(t *testing.T, serverURL string) *config.YunikornConfig {
-	parsedURL, err := url.Parse(serverURL)
-	require.NoError(t, err)
-
-	portNum, err := strconv.Atoi(parsedURL.Port())
-	require.NoError(t, err)
-
-	return &config.YunikornConfig{
-		Host: parsedURL.Hostname(),
-		Port: portNum,
-	}
-}
-
-func Test_GetApplications(t *testing.T) {
+func TestRESTClient_GetApplications(t *testing.T) {
 	tests := []struct {
 		name      string
 		partName  string
@@ -226,17 +213,7 @@ func Test_GetApplications(t *testing.T) {
 			}))
 
 			defer ts.Close()
-			serverURL, err := url.Parse(ts.URL)
-			require.NoError(t, err)
-
-			portNum, err := strconv.Atoi(serverURL.Port())
-			require.NoError(t, err)
-
-			cfg := config.YunikornConfig{
-				Host: serverURL.Hostname(),
-				Port: portNum,
-			}
-			client := NewRESTClient(&cfg)
+			client := NewRESTClient(getMockServerYunikornConfig(t, ts.URL))
 
 			apps, err := client.GetApplications(context.Background(), tt.partName, tt.queueName)
 			require.NoError(t, err)
@@ -248,7 +225,7 @@ func Test_GetApplications(t *testing.T) {
 	}
 }
 
-func Test_GetPartitions(t *testing.T) {
+func TestRESTClient_GetPartitions(t *testing.T) {
 	tests := []struct {
 		name     string
 		setup    func() *httptest.Server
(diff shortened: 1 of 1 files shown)


richscott (approved):

internal/yunikorn/rest_test.go
```diff
@@ -126,6 +127,616 @@ func TestRESTClient_GetApplication(t *testing.T) {
...
+					http.Error(w, err.Error(), http.StatusInternalServerError)
+				}
+
+			}))
+
+			defer ts.Close()
+			client := NewRESTClient(getMockServerYunikornConfig(t, ts.URL))
+
+			apps, err := client.GetApplications(context.Background(), tt.partName, tt.queueName)
+			require.NoError(t, err)
```
dejanzele, line 219: there is no sense doing further asserts in case of an error, all of them will fail.
```
if err != nil {
  t.Fatalf("error happened: %v", err)
}
```

internal/yunikorn/rest_test.go
```diff
@@ -126,6 +127,616 @@ func TestRESTClient_GetApplication(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ts := tt.setup()
+			defer ts.Close()
+
+			client := NewRESTClient(getMockServerYunikornConfig(t, ts.URL))
+
+			queues, err := client.GetPartitionQueues(context.Background(), "testPartition")
+			if tt.wantErr {
+				require.Error(t, err)
```
dejanzele, line 373: Can you also assert the error message is what we expect?

internal/yunikorn/rest_test.go
sudiptob2: Hi, I think further asserts only execute if there is no error.
`require.NoError(t, err) // If error it exists here`

richscott: Correct - any checks using `require` immediately stops the test (whereas `assert` checks will spew an error and keep going).

sudiptob2: yeah, so it's equivalent to `t.Fatalf()` anyway

internal/yunikorn/rest_test.go
sudiptob2: added

internal/yunikorn/rest_test.go
```diff
@@ -126,6 +127,640 @@ func TestRESTClient_GetApplication(t *testing.T) {
...
+			expectedErrMsg: "yunicorn api returned non-OK status code: 500",
+		},
+		{
+			name: "Unexpected JSON",
+			setup: func() *httptest.Server {
+				return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+					type unexpected struct {
+						Unexpected string `json:"unexpected"`
+					}
+					err := json.NewEncoder(w).Encode(&unexpected{Unexpected: "unexpected"})
```
dejanzele, line 444:
```suggestion
func write(t *testing.T, response any) {
  err := json.NewEncoder(w).Encode(response)
  if err != nil {
    t.Errorf("error writing response: %v", err)
  }
}
```

An error should never appear here, if it does, we are probably trying to return some malformed or invalid response and we want to fail.

sudiptob2: Hi, I could not understand this change properly. 
Yes, it will not return any error, because we are encoding a valid JSON. The purpose of this test was to test what happens if we return any JSON structure that is not `[]*dao.PartitionInfo` (for example)
As this is a test server setup, I was wondering if it still matters what error response we write.

sudiptob2: extracted repeated logic into `writeResponse()`

internal/yunikorn/rest_test.go
```diff
@@ -773,3 +737,10 @@ func getMockServerYunikornConfig(t *testing.T, serverURL string) *config.Yunikor
 		Port: portNum,
 	}
 }
+
+func writeResponse(w http.ResponseWriter, response any) {
+	err := json.NewEncoder(w).Encode(response)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
dejanzele, line 744: We always send valid responses, so we expect the error to be nil in all cases.

internal/yunikorn/rest_test.go
```diff
@@ -773,3 +737,10 @@ func getMockServerYunikornConfig(t *testing.T, serverURL string) *config.Yunikor
 		Port: portNum,
 	}
 }
+
+func writeResponse(w http.ResponseWriter, response any) {
```
dejanzele, line 741:
```suggestion
func writeResponse(t *testing.T, w http.ResponseWriter, response any) {
```

internal/yunikorn/rest_test.go
dejanzele, line 744:
```suggestion
		t.Fatalf("error writing response: %v", err)
```

internal/yunikorn/rest_test.go
```diff
@@ -773,3 +737,10 @@ func getMockServerYunikornConfig(t *testing.T, serverURL string) *config.Yunikor
...
+	err := json.NewEncoder(w).Encode(response)
```
dejanzele, line 742: Also you can add a `t.Helper()` at the start of the method so reported errors get reported from the parent test case.

internal/yunikorn/rest_test.go
sudiptob2: updated

internal/yunikorn/rest_test.go
sudiptob2: updated

dejanzele (approved):

PR #67, opened 2024-06-24
Update print statement with log statement.

### LINKED ISSUE #66: Update all print statement into log statement
We have implemented logging functionality using `zap` in #52. Now we need to convert all print statements using logger.

sudiptob2:
closes #66

### DIFF
--- a/cmd/event-collector/main.go
+++ b/cmd/event-collector/main.go
@@ -68,7 +68,7 @@ func main() {
 
 	k, err := loadConfig(cfgFile)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		log.Logger.Error(fmt.Sprintf("Error: %v", err))
 		os.Exit(1)
 	}
 
@@ -114,7 +114,7 @@ func main() {
 
 	repo, err := repository.NewECRepo(ctx, &ecConfig)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "could not create db repository: %v\n", err)
+		log.Logger.Error(fmt.Sprintf("could not create db repository: %v", err))
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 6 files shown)


cmd/event-collector/main.go
```diff
@@ -68,7 +68,7 @@ func main() {
 
 	k, err := loadConfig(cfgFile)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		log.Logger.Error(fmt.Sprintf("Error: %v", err))
```
dejanzele, line 71: Use either `Errorf` or `Errorw` depending how you want to include the error message, don't use `Sprintf` to format the message

cmd/event-collector/main.go
```diff
@@ -114,7 +114,7 @@ func main() {
 
 	repo, err := repository.NewECRepo(ctx, &ecConfig)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "could not create db repository: %v\n", err)
+		log.Logger.Error(fmt.Sprintf("could not create db repository: %v", err))
 		os.Exit(1)
```
dejanzele, line 118: this should be a `panic(err)`

cmd/event-collector/main.go
dejanzele: Also no need to have `Error:` in the message, it will be logger as error already

dejanzele: this applies to the rest also

dejanzele: and use lowercase messages, I see lowercase everywhere (don't capitalise)

sudiptob2: Hi, I tried the above suggestion but am having some problems. `log.Logger.Error()` receives a string whereas, `fmt.Errorf()` returns an `Error`. In that case, how can we use the logger?

dejanzele (approved):

Issue #63, opened 2024-06-17
Ensure resilience and make sure YHS does not miss events


dejanzele:
When upserting multiple values, should we terminate if we get an error in the middle?
What if database is temporarily unavailable?

Events should not be pieced together in memory but saves as soon as YHS received them. This will also remove statefulness from the application.

YHS should not crash if Yunicorn is not available.

PR #76, opened 2024-07-08
feat: add workqueue processing with exponential backoff when persisting data

dejanzele:
Closes https://github.com/G-Research/yunikorn-history-server/issues/63


internal/yunikorn/service.go
```diff
@@ -22,22 +23,28 @@ type Service struct {
...
 	}
 	s.eventHandler = s.handleEvent
+	go s.workqueue.Run(ctx)
+
 	return s
 }
 
 func (s *Service) Shutdown() {
 	s.eventCollectorRunning = false
+	s.workqueue.Shutdown()
```
richscott, line 47: Shouldn't this be before the previous line (`s.eventCollectorRunning = false`)?  Are we checking `evenCollectorRunning` somewhere far away?

internal/yunikorn/stream.go
```diff
@@ -64,20 +64,16 @@ func (s *Service) processStreamResponse(ctx context.Context, response []byte) er
 		logger.Errorf("error recording event: %v", err)
 	}
 
-	if eventRecord.GetType() == si.EventRecord_APP {
-		printAppEvent(&eventRecord)
-	}
+	logger.Infow(
```
richscott, line 67: Without the check for it being an App event, this can generate a huge amount of logging. I wonder if we should add a config parameter to allow filtering by event type, so we don't have all-or-nothing logging?

dejanzele: I don't have an opinion on this, you know more on this topic

richscott: Let's go with your change for now; I'll write a ticket to make filtering of event-logging configurable, but it doesn't need to be done now.

internal/yunikorn/service.go
dejanzele: refactored a bit

richscott (approved):

PR #90, opened 2024-07-11
Adds web remote modules

dcoric:
This pull request adds a module federated web app to the `web` directory. This new Angular project will be extended in the later stages with custom components for the YuniKorn. The web project is configured to output a `remoteEntry.js` file that will be loaded as a remote module to the host app (YuniKorn web).
The make file is extended with the `web-build` command that will output static files in the `assets` folder.


web/.gitignore
```diff
@@ -0,0 +1,42 @@
+# See http://help.github.com/ignore-files/ for more about ignoring files.
```
dejanzele, line 1: Should this be moved to the top-level .gitignore?

Makefile
```diff
@@ -347,3 +347,7 @@ KIND_VERSION ?= v0.23.0
 kind: $(KIND) ## Download kind locally if necessary.
 $(KIND): bin/tooling
 	test -s $(KIND) || GOBIN=$(LOCALBIN_TOOLING) $(GO) install sigs.k8s.io/kind@$(KIND_VERSION)
+
+.PHONY: web-build
+web-build: ## Build the web components
+	cd web && npm run build
```
dejanzele, line 353: is it possible to use `npm --prefix web` to we can run this command from the project root? This would help when chaining multiple make targets together

web/.gitignore
dcoric: I did some updates and merged it with the existing .gitignore

Makefile
dcoric: Thanks. I have updated it

dejanzele (approved):

PR #100, opened 2024-07-13
Implement test to verify queue creation

### LINKED ISSUE #80: Add integration test to verify creation of random queues

sudiptob2:
Closes #80 

Implement an end-to-end test to verify that YHS can observe queue creation events in the `yunikorn` scheduler. Also, verify that the event is saved and the data is visible in the web service.

**Definition of Done (DoD)**

- [x] Apply the config map that contains the queue definition. Reference: [partition-and-queue-configuration](https://yunikorn.apache.org/docs/1.5.0/user_guide/use_cases/#partition-and-queue-configuration).
- [x] Verify the `QUEUE-ADD` event counts.
- [x] Verify that the newly created queue is returned by the web service at the `/ws/v1/partition/:partition/queues` endpoint.

**Refactor**

- [x] Use `TestMain` to run the application once for all the tests.

**Notes**

Processing and returning the newly created queues by the web service takes considerable time. Therefore, the timeout is set to `400` seconds, making the process quite slow.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/G-Research/yunikorn-history-server/internal/webservice"
 	"github.com/G-Research/yunikorn-history-server/internal/yunikorn/model"
 	"github.com/G-Research/yunikorn-history-server/test/k8s"
+	"github.com/G-Research/yunikorn-history-server/test/util"
 	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/assert"
 	batchv1 "k8s.io/api/batch/v1"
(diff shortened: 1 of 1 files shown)


test/e2e/e2e_test.go
```diff
@@ -79,6 +95,69 @@ func TestYunikornEventStream_E2E(t *testing.T) {
 	}, 100*time.Second, 5*time.Second)
 }
 
+func TestYunikornQueueCreation_E2E(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping e2e test in short mode")
+	}
+	ns := "yunikorn"
+	queueName := "tenant" // a random name
```
richscott, line 103: What do you think about generating a small random string here, instead of a fixed-value? That way, there's no chance (or at least, a very small chance) that if this test func is run more than once against the same YK instance, there won't be this queue that was made in a previous run.  Maybe using the first, basic solution in `https://stackoverflow.com/questions/22892120/how-to-generate-a-random-string-of-a-fixed-length-in-go` ?

richscott (commented):
I had just one suggestion for randomizing the new queue name.

test/e2e/e2e_test.go
richscott: I see there's an existing `GenerateRandomAlphanum(t *testing.T, length int) string)` func in YHS - feel free to use that.

sudiptob2: updated

richscott (approved):

PR #108, opened 2024-07-17
Integration test fixes

richscott:
Remove application history check - this is inherently a flaky test, due to Yunikorn having very variable processing time for application history data, so we cannot rely on it without having a very long timeout value.

Fix check for expected error message when we specify a bad health endpoint URL - Linux gives a different trailing message after the "lookup invalid-host" substring. This now passes on both MacOS and Linux systems.

Make check for getting parititons just log the error, and not be fatal.

Fixes https://github.com/G-Research/yunikorn-history-server/issues/97

### DIFF
--- a/internal/health/health_int_test.go
+++ b/internal/health/health_int_test.go
@@ -44,8 +44,7 @@ func TestService_Readiness_Integration(t *testing.T) {
 		status := service.Readiness(context.Background())
 		assert.False(t, status.Healthy)
 		assert.Equal(t, 2, len(status.ComponentStatuses))
-		expectedError := "Get \"http://invalid-host:2212/ws/v1/scheduler/healthcheck\": dial tcp: lookup invalid-host: no such host"
-		assertStatus(t, status.ComponentStatuses, "yunikorn", false, expectedError)
+		assertStatus(t, status.ComponentStatuses, "yunikorn", false)
 		assert.Equal(t, now, status.StartedAt)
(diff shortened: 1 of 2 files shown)


dejanzele (approved):

internal/health/health_int_test.go
```diff
@@ -75,11 +74,13 @@ func TestService_Readiness_Integration(t *testing.T) {
 	})
 }
 
-func assertStatus(t *testing.T, statuses []*ComponentStatus, identifier string, expectedHealthy bool, expectedError string) {
+func assertStatus(t *testing.T, statuses []*ComponentStatus, identifier string, expectedHealthy bool) {
+	expectedErrorPrefix := `Get "http://invalid-host:2212/ws/v1/scheduler/healthcheck": dial tcp: lookup invalid-host`
```
dejanzele, line 78: I think we should provide the error to the func as we might simulate in the future other errors.
I'd define it in the test case as `wantErrorMsg` and assert it here it contains the error message.

dejanzele (commented):
One comment regarding hardcoded error prefix

internal/health/health_int_test.go
richscott: > I think we should provide the error to the func as we might simulate in the future other errors. I'd define it in the test case as `wantErrorMsg` and assert it here it contains the error message.

Good suggestion - fixed in commit cf4e4dd

dejanzele (approved):

Issue #71, opened 2024-07-05
YHS event-collector always requires a config file to run without defaulting to settings from env


richscott:
The YHS event-collector now (commit 2969bd9) always requires a config file to run, and exits with an error if one if not specified - it should fall back to using values from the environment if a config file is not given.

PR #113, opened 2024-07-19
feat: do not require a config file when starting YHS

dejanzele:
Closes https://github.com/G-Research/yunikorn-history-server/issues/71


internal/config/config_test.go
```diff
@@ -1,8 +1,11 @@
 package config
 
 import (
```
richscott, line 3: This doesn't break anything, but don't some/most/all IDEs sort imports in the same order as `goimports`? (stdlib packages first, then public/third-party ones, etc)

richscott (dismissed):

internal/config/config_test.go
sudiptob2: I am using `goimports` in GoLand which sort that way.

dejanzele: linter was broken, I pushed a fix and needed to resolve various lint issues

richscott: 👏🏼

richscott (approved):

PR #118, opened 2024-07-23
publish helm chart only on tags

ljubon:
As discussed internally, proposed way to publish helm chart should be only when pushing tags starting with `refs/tags/v`


pavlovic-ivan (approved):
LGTM. Thanks @ljubon! 🚀

Issue #99, opened 2024-07-12
Add query parameters to YHS data endpoints for filtering


richscott:
As YHS is intended to support several of the same endpoints as the Yunikorn Scheduler's API, with added enhancement of being able to retrieve historical data that it stores, we need to add query parameters to several endpoints to allow the user to filter the queried results by date range, queue, etc.

Initially, handlers for the following endpoints should have date-range filter parameter logic added:

- [ ] `/ws/v1/partition/:partition_name/queues`
- [ ] `/ws/v1/partition/:partition_name/queue/:queue_name/applications`
- [ ] `/ws/v1/history/apps`
- [ ] `/ws/v1/history/containers`

Filtering parameters should be optional. They are:
- `start=nnnn` - records should have creation time (e.g. `submission_time` for applications) of this date or after. The `nnnn` value is an integer representing the the number of milliseconds for this date since the "epoch", defined as the midnight at the beginning of January 1, 1970, UTC.
- `end=nnnn` - records should have creation time no later than this value (the same units as `start` above) (_TODO: should this be queried as `finished_time` for Applications?_)  If this parameter is not specified at all, all records with creation up to the present moment should be returned.
- `offset=N` - for paginated output, this should be the ordinal number of the first record found in the YHS database for this query. If not specified this should default to 0.
- `limit=N` - for paginated output, this is the number of records to return in this response. If not specified, it should default to a system-configurable value, perhaps 500.
- `user=job-submitter-user` - constrain the results set to only those submitted by that one user (only one username can be used)
- `groups=group1,group2,group3` - filter result set to include only results that are in *any* of the specified groups

PR #129, opened 2024-07-29
feat: add support for query params for application endpoints

dejanzele:
Closes https://github.com/G-Research/yunikorn-history-server/issues/99


internal/webservice/helpers.go
```diff
@@ -0,0 +1,123 @@
...
+	queryParamOffset              = "offset"
+	queryParamUser                = "user"
+)
+
+func parseApplicationFilters(r *http.Request) (*repository.ApplicationFilters, error) {
+	filters := repository.ApplicationFilters{}
+	user := getUserQueryParam(r)
+	if user != "" {
+		filters.User = &user
+	}
```
richscott, line 27: Do we need to add/set a filter for `groups` here? I don't see any calls to `getGroupsQueryParam(..)` around here.

internal/webservice/response.go
```diff
@@ -49,3 +49,19 @@ func errorResponse(w http.ResponseWriter, r *http.Request, err error) {
 		log.Logger.Errorf("could not write error response: %v", err)
 	}
 }
+
+func badRequestResponse(w http.ResponseWriter, r *http.Request, err error) {
+	log.Logger.Errorf("error processing request for %s: %v", r.URL.Path, err)
+	problemDetails := ProblemDetails{
+		Type:     "about:blank",
```
richscott, line 56: I always thought it was silly that the RFC specifies this field, instead of just omitting it - this Is JSON, for god's sake!

internal/database/repository/application_int_test.go
```diff
@@ -0,0 +1,268 @@
...
+		{
+			ApplicationID:   "app1",
+			UsedResource:    map[string]int64{"cpu": 1},
+			MaxUsedResource: map[string]int64{"cpu": 2},
+			PendingResource: map[string]int64{"cpu": 1},
+			Partition:       "default",
+			QueueName:       "root.default",
+			SubmissionTime:  now.Add(-6 * time.Hour).UnixMilli(),
+			User:            "user1",
+			State:           "RUNNING",
```
richscott, line 199: What do you think about putting all the states in a set of string constants? Or maybe yunikorn already exports them?

richscott: I just looked at the `yunikorn-scheduler-interface` proto definitions - they define state changes like `EventRecord_APP_RUNNING` - see `lib/go/si/si.pb.go`.  Think it's worth it to use those?

internal/util/util.go
```diff
@@ -8,3 +12,23 @@ func ToPtrSlice[T any](slice []T) []*T {
 	}
 	return result
 }
+
+func ToPtr[T any](x T) *T {
+	return &x
+}
+
+// SliceToCommaSeparated converts a string slice to a comma-separated string.
+func SliceToCommaSeparated(slice []string) string {
```
richscott, line 21: Would it be simpler for the below code to use `strings.Join()` on single-quoted values? Maybe something like
```
       mySlice := []string{"dejan", "rich", "dave", "denis"}
        quoted := []string{}
        for _, s := range mySlice {
                quoted = append(quoted, "'"+s+"'")
        }
        commastr := strings.Join(quoted, ",")
```

richscott (commented):
Just a few more observations for now -  I will look again.

internal/webservice/helpers.go
dejanzele: added

internal/database/repository/application.go
```diff
@@ -3,12 +3,45 @@ package repository
...
+	FinishedEndTime     *time.Time
+	User                *string
+	Offset              *int
+	Limit               *int
+}
+
+// applyApplicationFilters adds application filters to the sql query.
+func applyApplicationFilters(builder *sql.Builder, filters ApplicationFilters) {
+	if filters.SubmissionStartTime != nil {
+		builder.Conditionf("submission_time >= '%d'", filters.SubmissionStartTime.UnixMilli())
```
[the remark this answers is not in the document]
dejanzele: added

internal/util/util.go
dejanzele: added 2 modes, one with quotes and one without. In SQL we need them quoted.

internal/database/repository/application_int_test.go
dejanzele: addressed

internal/util/util.go
```diff
@@ -8,3 +12,27 @@ func ToPtrSlice[T any](slice []T) []*T {
...
+
+// SliceToCommaSeparated converts a string slice to a comma-separated string.
+// If quoteItems is true, each item in the slice will be quoted.
+func SliceToCommaSeparated(slice []string, quoteItems bool) string {
+	if len(slice) == 0 {
+		return ""
+	}
+	if !quoteItems {
+		return strings.Join(slice, ",")
+	}
```
richscott, line 28: You could use `strings.Join()` for both cases, and it would be even shorter, e.g.
```
func SliceToCommaSeparated(slice []string, quoteItems bool) string {
        if quoteItems {
                quoted := []string{}
                for _, s := range slice {
                        quoted = append(quoted, "'"+s+"'")
                }
                return strings.Join(quoted, ",")
        }

        return strings.Join(slice, ",")
}
```
This will work correctly (for quoting and adding commas) on slices having 0, 1, or >1 elements.

richscott (commented):
Just one more suggested refinement of `SliceToCommaSeparated()` - it could be even shorter...

internal/util/util.go
dejanzele: addressed

richscott (approved):

PR #130, opened 2024-08-01
feat: make data sync interval configurable

dejanzele:


internal/yunikorn/service.go
```diff
@@ -29,7 +29,15 @@ type Service struct {
...
-func NewService(ctx context.Context, repository repository.Repository, eventRepository repository.EventRepository, client Client) *Service {
+type Option func(*Service)
+
+func WithSyncInterval(interval time.Duration) Option {
+	return func(s *Service) {
+		s.syncInterval = interval
+	}
+}
+
+func NewService(repository repository.Repository, eventRepository repository.EventRepository, client Client, opts ...Option) *Service {
```
richscott, line 40: `ctx` is deleted - do/will we need a Context for doing cancellation, etc, in the future?

richscott (commented):
Just one question about removing a Context in the NewService() constructor.

internal/yunikorn/service.go
dejanzele: the context was actually unused, it got changed now to start a goroutine during instantiate, so it was just sitting idle

richscott: 👍🏼

richscott (approved):

Issue #134, opened 2024-08-05
Use `minikube` as alternative to `kind` for testing


richscott:
When running `make test` on YHS in the C3 environment, `kind` consistently hangs at the "Starting control-plane" message. Armada devs have mentioned they cannot use *kind* in their environment. Spark/K8S devs have mentioned using `minikube` for their testing. We should try providing support for using `minikube` for testing YHS as a switchable alternative to `kind`.

PR #135, opened 2024-08-06
Add support for using `minikube` as cluster manager

richscott:
These changes add support to running the tests with using `minikube` as the K8S cluster manager as an alternative to `kind`, which may not work in some environments.

The default cluster manager is still `kind` when running tests, but minikube can be used by doing:
```
env CLUSTER_MGR=minikube make test
```

Also, the `helm` command is now downloaded to `bin/tooling/` and used from there, for any environments that may not already have it installed. 

The `kind-create-cluster` and `kind-delete-cluster` targets are now renamed to simply `create-cluster` and `delete-cluster`. 

Fixes https://github.com/G-Research/yunikorn-history-server/issues/134

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -368,25 +368,26 @@ install-and-patch-yunikorn: helm-install-yunikorn patch-yunikorn-service ## inst
 
 .PHONY: helm-install-yunikorn
 helm-install-yunikorn: ## install yunikorn using helm.
-	helm upgrade --install yunikorn yunikorn/yunikorn --namespace $(NAMESPACE) --create-namespace
+	$(HELM) upgrade --install yunikorn yunikorn/yunikorn --namespace $(NAMESPACE) --create-namespace
 
 .PHONY: helm-uninstall-yunikorn
 helm-uninstall-yunikorn: ## uninstall yunikorn using helm.
-	helm uninstall yunikorn --namespace $(NAMESPACE)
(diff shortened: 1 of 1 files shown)


sudiptob2 (approved):
Do we need to update the `Readme`.  Specially the testing section?

Makefile
```diff
@@ -87,9 +87,16 @@ BASE_IMAGE ?= alpine
 BASE_IMAGE_TAG ?= 3.20
 
 # Local Development
-KIND_CLUSTER ?= yhs
+CLUSTER_MGR ?= kind     # either 'kind' or 'minikube'
+CLUSTER_NAME ?= yhs
 NAMESPACE ?= yunikorn
 
+KIND ?= $(LOCALBIN_TOOLING)/kind
+KIND_VERSION ?= v0.23.0
```
dave-gantenbein, line 95: Why not "latest"?

richscott:
> Do we need to update the `Readme`. Specially the testing section?

Yes - fixed in commit 672c721

Makefile
richscott: Good suggestion - it was at that fixed version before, but it's safe to use 'latest' - kind is very heavily tested before every release. Fixed in commit 672c721.

dave-gantenbein (approved):
lgtm

PR #149, opened 2024-08-14
Modify queues DB table to separate out all child queues

### LINKED ISSUE #142: Modify queues DB table to separate out all child queues as separate rows
Currently, the `queues` table in the YHS database does not lend itself to easily querying for Yunikorn's hierarchical queues. When YHS sees a queue-creation event, it inserts a new row into the `queues` table with all the usual fields (queue_name, status, partition, etc), and we store that queue's child queues all in a single field - `children` of type `JSONB`; we also store another field `children_names`, of type `TEXT[]`. This makes searching for a queue and all its descendants complex and slow - we have to fetch the column, convert the JSON field to a Golang tree of queues, and walk down the tree.

We should change the data-model to represent each queue as a single-row, with no children encapsulated in that row, and have every child queue also be its own row. We can use the SQL design of representing parent-child relationships within the same table as described in https://learnsql.com/blog/query-parent-child-tree/ (as well as many other articles).

This ticket is about just changing the queues table - other tickets will describe the db insert process and queries. The changes for the `queues` table should be:

- [ ] remove `children JSONB` column
- [ ] remove `children_names TEXT[]` column
- [ ] add a `parent_id UUID` column - should put it between the id and queue_name columns
- [ ] add a `created BIGINT NOT NULL` column
- [ ] add a `deleted BIGINT` column (needs to be nullable, obviously)

Also, the `applications` table should be modified:
- [ ] add a `queue_id UUID NOT NULL`  column, between `app_id` and `used_resource`
- [ ] OPTIONAL:  maybe remove the `queue_name`? But if there are many queries that won't need to join against the queues table for more info, maybe we still want this.

sudiptob2:
closes #142


internal/database/repository/application.go
```diff
@@ -51,10 +51,10 @@ func applyApplicationFilters(builder *sql.Builder, filters ApplicationFilters) {
 
 func (s *PostgresRepository) UpsertApplications(ctx context.Context, apps []*dao.ApplicationDAOInfo) error {
 	upsertSQL := `INSERT INTO applications (id, app_id, used_resource, max_used_resource, pending_resource,
-			partition, queue_name, submission_time, finished_time, requests, allocations, state,
+			partition, queue_name, queue_id, submission_time, finished_time, requests, allocations, state,
```
richscott, line 54: What's the difference between `queue_id` and the existing `id` column? Do they need to be separate?

sudiptob2: existing `id` is the id column for the application table, queue_id is `id` column of the `queue` table

migrations/20240628155246_init-tables.up.sql
```diff
@@ -62,13 +64,13 @@ CREATE TABLE queues(
...
     parent TEXT,
     template_info JSONB,
-    children JSONB,
-    children_names TEXT[],
     abs_used_capacity JSONB,
     max_running_apps INTEGER,
     running_apps INTEGER NOT NULL,
     current_priority INTEGER,
     allocating_accepted_apps TEXT[],
+    created_at BIGINT NOT NULL,
```
richscott, line 72: Can you move these two new columns up, so they are right after `parent_id` (and before `queue_name`)  - I expect we will have a number of queries against these, and there is a slight performance improvement when a column has only fixed-width-type columns before it (no TEXT or JSONB, etc).

richscott (commented):
Just one suggestion on the queues table changes in the db migration.

richscott (approved):

Issue #143, opened 2024-08-12
Rework queues DB logic for separate CRUD functions and queue entries


richscott:
After https://github.com/G-Research/yunikorn-history-server/issues/142 has been implemented, replace the  `UpsertQueues(...)` function with separate "CRUD"-style functions, so we can correctly store and process the hierarchical model of Yunikorn queues.

- [x] https://github.com/G-Research/yunikorn-history-server/issues/181
- [x] https://github.com/G-Research/yunikorn-history-server/issues/182 
- [x] https://github.com/G-Research/yunikorn-history-server/issues/183
- [x] https://github.com/G-Research/yunikorn-history-server/issues/184
- [x] https://github.com/G-Research/yunikorn-history-server/issues/185
- [x] https://github.com/G-Research/yunikorn-history-server/issues/186

Issue #148, opened 2024-08-14
Add PartitionName for child queues


sudiptob2:
When we retrieve queue data from the YuniKorn API `(/ws/v1/partition/default/queues)`, the `partitionName` is only provided for the top level queue and `partitionName` field is empty (`""`) for these child queues. 
Therefore, In YHS, when we store all child queues as individual rows; we also store empty string for `partition_name`.

This partition name is crucial when we query based on `queueName`.

<img width="594" alt="Screenshot 2024-08-14 at 10 50 51 AM" src="https://github.com/user-attachments/assets/eae9eac1-aa2d-46bc-a4e4-3731a29a864f">

**Possible  Solution:**  
While flattening the queues in the **recursive** `flattenQueues()` method, inject the partition name from the top-level queue into each of the child queues.
https://github.com/G-Research/yunikorn-history-server/blob/91580b04d7acc97e0234858f0dfa564769f74696/internal/yunikorn/sync.go#L149-L158

PR #155, opened 2024-08-20
Implement `AddQueues()` - recursively insert a queue and all its children directly.

richscott:
This is so we don't have to expect callers to "flatten" their queues, nor have them populate parent_id values for descendants of a top-level queue.

Also, update the Makefile to consistently use Uber's "mockgen" utility instead of the deprecated "gomock" tooling, and regenerate db repository interface mocks.

Fixes https://github.com/G-Research/yunikorn-history-server/issues/181
related to #143


internal/database/repository/queue.go
```diff
@@ -51,10 +53,14 @@ func (s *PostgresRepository) UpsertQueues(ctx context.Context, queues []*dao.Par
 		is_managed = EXCLUDED.is_managed,
 		max_running_apps = EXCLUDED.max_running_apps,
 		running_apps = EXCLUDED.running_apps`
+
 	for _, q := range queues {
-		parentId, err := s.getQueueID(ctx, q.Parent, q.Partition)
-		if err != nil {
-			return fmt.Errorf("could not get parent queue from DB: %v", err)
+		var err error
```
dave-gantenbein, line 58: Is the expected number of queues always going to be small? Asking for a friend who thinks this will query N times per upsert.

internal/database/repository/queue.go
```diff
@@ -82,7 +88,18 @@ func (s *PostgresRepository) UpsertQueues(ctx context.Context, queues []*dao.Par
...
+		}
+
+		if len(q.Children) > 0 {
+			children := []*dao.PartitionQueueDAOInfo{}
+			for _, child := range q.Children {
+				children = append(children, &child)
+			}
+			err = s.UpsertQueues(ctx, parentId, children)
+			if err != nil {
+				return fmt.Errorf("could not one or more children of queue %s into DB: %v", q.QueueName, err)
```
sudiptob2, line 101:
```suggestion
				return fmt.Errorf("could not upsert/insert one or more children of queue %s into DB: %v", q.QueueName, err)
```

internal/yunikorn/sync.go
```diff
@@ -131,10 +131,10 @@ func (s *Service) upsertPartitionQueues(ctx context.Context, partitions []*dao.P
 		return nil, fmt.Errorf("failed to get queues for some partitions: %v", errs)
 	}
 
-	queues = flattenQueues(queues)
+	// queues = flattenQueues(queues)
```
sudiptob2, line 134: Remove this line?

sudiptob2: Also probably we can remove the `flattenQueues()` ?

sudiptob2 (commented):
I tested the branch locally. At the current state it is not working. I guess following pice of the logic is missing in the implementation. 
```GO
func flattenQueues(qs []*dao.PartitionQueueDAOInfo) []*dao.PartitionQueueDAOInfo {
	var queues []*dao.PartitionQueueDAOInfo
	for _, q := range qs {
		queues = append(queues, q)
		if len(q.Children) > 0 {
			// update partitionName for children #148
			for i := range q.Children {
				q.Children[i].Partition = q.Partition
			}
			queues = append(queues, flattenQueues(util.ToPtrSlice(q.Children))...)
		}
	}
	return queues
}
```

In #148, we added partitionName to the child queues because yunikorn-core provides an empty string "" for partitionName. Since we're removing flattenQueues in this implementation, we need to address this logic elsewhere. If not, getParentID() won’t return the correct parentId for the queues.

internal/database/repository/queue.go
```diff
@@ -82,7 +88,18 @@ func (s *PostgresRepository) UpsertQueues(ctx context.Context, queues []*dao.Par
...
+			return fmt.Errorf("could not insert/update queue %s into DB: %v", q.QueueName, err)
+		}
+
+		if len(q.Children) > 0 {
+			children := []*dao.PartitionQueueDAOInfo{}
+			for _, child := range q.Children {
+				children = append(children, &child)
+			}
+			err = s.UpsertQueues(ctx, parentId, children)
+			if err != nil {
```
sudiptob2, line 100: I might be mistaken, but shouldn't the recursion be called with the current `queueId` as the parentID for the children? Otherwise, it seems like the `rootQueue` will be set as the parent for all generations of the queues.

internal/database/repository/queue.go
sudiptob2: @dave-gantenbein 
Hmm... we might need to think this through a bit more. Given that the queues are organized in a parent-child structure and can be nested down to `D` levels, we need to insert the parent first to allow for inserting its children. While we could potentially insert all the direct children in a single query, in that case, time complexity will be `O(D)`, where `D` is the depth of the tree. This will be a slight improvement but the worst case will still be `O(N)` where `N==D` for a skewed tree
`N = Number of nodes in the tree`

internal/database/repository/queue.go
dejanzele: upsert is a combination of insert and update

Makefile
```diff
@@ -145,8 +145,8 @@ migrate-down: ## migrate down using gomigrate.
 ##@ Codegen
 
 .PHONY: codegen
-codegen: gomock ## generate code using go generate (mocks).
-	go generate ./...
+codegen: mockgen ## generate code using go generate (mocks).
+	PATH=$(LOCALBIN_TOOLING):$$PATH go generate ./...
```
dejanzele, line 149: what does this do?

internal/database/repository/repository.go
```diff
@@ -27,7 +27,7 @@ type Repository interface {
 	GetNodesPerPartition(ctx context.Context, partition string) ([]*dao.NodeDAOInfo, error)
 	UpsertPartitions(ctx context.Context, partitions []*dao.PartitionInfo) error
 	GetAllPartitions(ctx context.Context) ([]*dao.PartitionInfo, error)
-	UpsertQueues(ctx context.Context, queues []*dao.PartitionQueueDAOInfo) error
+	UpsertQueues(ctx context.Context, parentId *string, queues []*dao.PartitionQueueDAOInfo) error
```
dejanzele, line 30: Are there any semantic differences if `parentId` is `null` opposed to `""`?

Makefile
richscott: This is to ensure we can run `bin/tooling/codegen` - `bin/tooling` is not in my default PATH, so Make wasn't finding `codegen` (I did have an old `gomock` in my `$HOME/go/bin`, but we want to be sure we are using `bin/tooling/` for whatever tools we download/install in this Make processing.

internal/database/repository/queue.go
```diff
@@ -12,7 +12,9 @@ import (
 	"github.com/G-Research/yunikorn-history-server/internal/model"
 )
 
-func (s *PostgresRepository) UpsertQueues(ctx context.Context, queues []*dao.PartitionQueueDAOInfo) error {
+func (s *PostgresRepository) UpsertQueues(ctx context.Context, parentId *string, queues []*dao.PartitionQueueDAOInfo) error {
+	//  XXX change ON CONFLICT clause to CONFLICT(id), since we may have the same queue+partition name over
+	// time in the same cluster?
```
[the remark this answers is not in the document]
richscott: But over a long period of time (days/weeks/months), we will surely encounter the same queue+partition combo in the table - just that some of those will have their `deleted_at` column set to some value. So, maybe we should change the `UNIQUE(id)` constraint on `queues` to `UNIQUE(queue_name, partition, deleted_at)` where `deleted_at` is non-null (there is a way in Postgresql to specify a unique constraint column has to have a non-null value).

internal/database/repository/queue.go
richscott: We could bundle insertions of child queues into single INSERT transactions (perhaps all siblings at each generation), and then downward - or all descendants of the top-level queue in one INSERT (or txn). But I'd be surprised if an organization would be creating more than hundreds of queues per day - I expect `applications`  will get the most traffic each day, by far. 

Note that in the changed code (lines 59-63), we call `getQueueID(..)` only if `UpsertQueues(..)` is not passed a given parentID - so the first/top-level call to `UpsertQueues()` will cause this to query for a queue ID, then all child queues will have this queue's id as its parentID, and so on down the descendants tree. There should be only one call to getQueueID() in the chain.

internal/database/repository/queue.go
richscott: You are exactly right! This is a bad typo. Fixed in commit 41f325a.

internal/database/repository/queue.go
richscott: Fixed in commit 4473aa6

internal/database/repository/repository.go
richscott: I made it a pointer to a string as `queues.parent_id` is nullable in the db, and @sudiptob2's `getQueueId()` already returned a `*string`.

internal/yunikorn/sync.go
richscott: Aha - just realized the only remaining real reference to `flattenQueues()` was its unit-test! Now removed in commit b6628d4.  Good catch.

internal/database/repository/queue.go
```diff
@@ -82,7 +90,18 @@ func (s *PostgresRepository) UpsertQueues(ctx context.Context, queues []*dao.Par
...
+
+		if len(q.Children) > 0 {
+			children := []*dao.PartitionQueueDAOInfo{}
+			for _, child := range q.Children {
+				children = append(children, &child)
+			}
+			err = s.UpsertQueues(ctx, &queueId, children)
+			if err != nil {
+				return fmt.Errorf("could not insert/update one or more children of queue %s into DB: %v", q.QueueName, err)
+			}
```
sudiptob2, line 104: This seems bit tricky. What happens when we do an update? I guess it will pass a newly generated ID from line no 66 to the child, which is incorrect.

Probably we need to use the `RETURNING` keyword in Postgres. Something like,
`INSERT INTO items (name) values ('bear') RETURNING id;`

Makefile
dejanzele: I think you can provide `GOBIN` to be equal to `LOCALBIN_TOOLING` and that should be considered first. (not 100% tho)

internal/database/repository/queue.go
richscott: I agree - fixed in commit af53844 .

internal/database/repository/repository.go
richscott: The `UpsertQueues()` is now deprecated (we will remove it when this PR, which adds `AddQueues()` and the PR for `UpdateQueue()` lands). Iif a pointer to an empty string var is passed in a queue struct to `AddQueues()`, the insert should fail, as a valid UUID is required for that column. (The db column is literally of type `UUID`.)

internal/yunikorn/sync.go
```diff
@@ -143,24 +141,6 @@ func (s *Service) upsertPartitionQueues(ctx context.Context, partitions []*dao.P
...
-			// update partitionName for children #148
-			for i := range q.Children {
-				q.Children[i].Partition = q.Partition
-			}
-			queues = append(queues, flattenQueues(util.ToPtrSlice(q.Children))...)
-		}
-	}
-	return queues
-}
-
```
sudiptob2, line 163: We need to revert it for now, as we are not modifying `UpsertQueue()` anymore. In later ticket we have to update the sync logic and replace `UpsertQueue()` with AddQueues() and UpdateQueues().

internal/database/repository/queue.go
```diff
@@ -88,6 +88,73 @@ func (s *PostgresRepository) UpsertQueues(ctx context.Context, queues []*dao.Par
 	return nil
 }
 
+func (s *PostgresRepository) AddQueues(ctx context.Context, parentId *string, queues []*dao.PartitionQueueDAOInfo) error {
+	upsertSQL := `INSERT INTO queues (
```
sudiptob2, line 92:
```suggestion
	insertSQL := `INSERT INTO queues (
```

sudiptob2 (changes requested):

sudiptob2:
Currently, we're encountering the following error in the `integration test`. The migration works fine when tested locally, but for some reason, the migrations don't sync during the test.

We can use `gen_random_uuid () → uuid`, which doesn't require any extensions to be installed and is commonly used in many applications.

Reference: [PostgreSQL UUID Functions](https://www.postgresql.org/docs/current/functions-uuid.html)

---
```
=== Failed
=== FAIL: internal/database/migrations TestGoMigrate_Integration (0.10s)
2024-08-30T21:12:07.373Z	INFO	created schema test_7fxse: CREATE SCHEMA
2024-08-30T21:12:07.373Z	INFO	set search path to test_7fxse: SET
2024-08-30T21:12:07.384Z	INFO	running migrate up
    gomigrate_int_test.go:33: 
        	Error Trace:	/home/runner/work/yunikorn-history-server/yunikorn-history-server/internal/database/migrations/gomigrate_int_test.go:33
        	Error:      	Should be true
        	Test:       	TestGoMigrate_Integration
        	Messages:   	expected up migrations to be applied for the first run
    gomigrate_int_test.go:35: error running migrations up: migration failed: function uuid_generate_v1() does not exist (column 30) in line 51: -- Create partitions table
        CREATE TABLE partitions(
            id UUID,
            cluster_id TEXT NOT NULL,
            name TEXT NOT NULL,
            capacity JSONB,
            used_capacity JSONB,
            utilization JSONB,
            total_nodes INTEGER,
            applications JSONB,
            total_containers INTEGER,
            state TEXT,
            last_state_transition_time BIGINT,
            UNIQUE (id),
            UNIQUE (name),
            PRIMARY KEY (id)
        );
```

dave-gantenbein (approved):

Makefile
sudiptob2: `GOBIN=$(LOCALBIN_TOOLING) $(GO) generate ./...` but still get path error.

sudiptob2 (approved):

Issue #165, opened 2024-08-24
[helm-chart installation] history-server installation with helm chart fail


0yukali0:
# What is this issue?
micro-service mode fails.

# How to reproduce?
```
make kind-all
helm install history-server chart -n yunikorn
```

# Screenshots
![image](https://github.com/user-attachments/assets/9d238c03-ff12-4675-9607-871109f951e6)
![image](https://github.com/user-attachments/assets/5d72a341-4f0d-475c-91fc-1c606bd9e797)

sudiptob2:
solved in #166

PR #166, opened 2024-08-24
[YHS-165] upadte build and add command in chart

0yukali0:
# What is PR?
This PR fixs the failure by adopts CGO option when compiling and adds the command to the chart/template/deployment.yml. 

# Related issues
#165 

# How to validate?
```
# prepare a kind cluster
make kind-all
# load custom image to the kind
make kind-load-image
sed -i 's/tag: .*/tag: "<YOUR IMAGE TAG>"/g' charts/yunikorn-history-server/values.yaml
helm install yhs chart -n yunikorn
```
# Screenshots
image
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -363,7 +363,7 @@ endif
 
 .PHONY: kind-load-image
 kind-load-image: docker-build-amd64 ## inject the local docker image into the kind cluster.
-	kind load docker-image $(IMAGE_TAG) --name $(KIND_CLUSTER)
+	kind load docker-image $(IMAGE_TAG) --name $(CLUSTER_NAME)
 
 .PHONY: install-dependencies
 install-dependencies: helm-repos install-and-patch-yunikorn helm-install-postgres wait-for-dependencies ## install dependencies.
--- a/Makefile
+++ b/Makefile
@@ -274,7 +274,7 @@ web-build: ng ## build the web components.
 .PHONY: build
 build: bin/app ## build the yunikorn-history-server binary for current OS and architecture.
 	echo "Building yunikorn-history-server binary for $(OS)/$(ARCH)"
-	GOOS=$(OS) GOARCH=$(ARCH) $(GO) build -o $(LOCALBIN_APP)/yunikorn-history-server 										\
+	CGO_ENABLED=0 GOOS=$(OS) GOARCH=$(ARCH) $(GO) build -o $(LOCALBIN_APP)/yunikorn-history-server 										\
 		-ldflags "-X github.com/G-Research/yunikorn-history-server/cmd/yunikorn-history-server/info.Version=$(GIT_TAG) 		\
 				  -X github.com/G-Research/yunikorn-history-server/cmd/yunikorn-history-server/info.Commit=$(GIT_COMMIT) 	\
 				  -X github.com/G-Research/yunikorn-history-server/cmd/yunikorn-history-server/info.BuildTime=$(BUILD_TIME)" \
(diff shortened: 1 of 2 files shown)


0yukali0:
#165

Makefile
```diff
@@ -274,7 +274,7 @@ web-build: ng ## build the web components.
 .PHONY: build
 build: bin/app ## build the yunikorn-history-server binary for current OS and architecture.
 	echo "Building yunikorn-history-server binary for $(OS)/$(ARCH)"
-	GOOS=$(OS) GOARCH=$(ARCH) $(GO) build -o $(LOCALBIN_APP)/yunikorn-history-server 										\
+	CGO_ENABLED=0 GOOS=$(OS) GOARCH=$(ARCH) $(GO) build -o $(LOCALBIN_APP)/yunikorn-history-server 										\
```
sudiptob2, line 277: Hi, What is the particular reason for disabling CGO? I don't think it is related to the failure. Could you provide a little bit more details please?

charts/yunikorn-history-server/templates/deployment.yaml
```diff
@@ -39,6 +39,7 @@ spec:
         - name: "yunikorn-history-server"
           image: "{{ include "yunikorn-history-server.image" . }}"
           imagePullPolicy: "{{ $imagePullPolicy }}"
+          command: ["/app/yunikorn-history-server"]
```
sudiptob2, line 42: We already have this command to the `Entrypoint` of our docker image. Therefore, we do not need this.
```suggestion
```

sudiptob2 (changes requested):
Hi, Thanks for the contribution 🚀 
I think the only problem with the helm deployment is wrong cluster name. The variable `KIND_CLUSTER` does not exist. Therefore we have to use `CLUSTER_NAME` variable.
Nice catch on that 🙏

Makefile
0yukali0: Hi @sudiptob2, thanks for review.
Binary file in the alpine is not executable without disabling CGO. After setting this option, it is executable.
Current version.
image
image
Modified version.
image
[Ref](https://stackoverflow.com/questions/36279253/go-compiled-binary-wont-run-in-an-alpine-docker-container-on-ubuntu-host)

charts/yunikorn-history-server/templates/deployment.yaml
0yukali0: Agree.
I will remove this in next commit.

Makefile
0yukali0: Hi @sudiptob2. 
My machine setting is following.
Description:    Ubuntu 22.04.3 LTS
Release:        22.04
Arch:  x86_64

richscott: @0yukali0 @sudiptob2 I tested on my Ubuntu 22.04.4 LTS system (x86_64 arch) and was seeing the error - it does not build the `bin/app/yunikorn-history-server` - the solution of using `CGO_ENABLED=0` does work. I found out from reading

  https://stackoverflow.com/questions/61515186/when-using-cgo-enabled-is-must-and-what-happens

that CGO is disabled by default when cross-compiling (which is why we didn't see it when doing our builds on macOS/arm64) but enabled by default on native compilations. I think this PR is good.

richscott (approved):

sudiptob2 (approved):
`CGO` isn't strictly required for us, so disabling it by default seems the most user-friendly option.

PR #168, opened 2024-08-26
Added context path for web

dcoric:
Adds context path for web as a parameter in Makefile `WEB_ROOT`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,9 @@ GIT_TAG ?= $(shell git describe --tags --dirty --always)
 # IMAGE_TAG defines the name and tag of the operator image.
 IMAGE_TAG ?= $(IMAGE_REPO):$(GIT_TAG)
 
+# WEB_ROOT defines path that will open web UI.
+WEB_ROOT ?= /web/
+
 # Go compiler selection
 GO := go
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -269,7 +272,7 @@ test-k6-performance: ## run k6 performance tests.
 
 .PHONY: web-build
 web-build: ng ## build the web components.
-	npm install --prefix web && npm run build --prefix web
+	npm install --prefix web && npm run build --prefix web --base-href $(WEB_ROOT)
```
sudiptob2, line 275:
```suggestion
	npm install --prefix web && npm run build --prefix web -- --project yhs-components --base-href $(WEB_ROOT)
```
Hi, Current `web-build` command gives an error. Probably we need to change it to something I suggested above?

image

sudiptob2 (changes requested):

Makefile
dcoric: Ah, yes, npm requires `--` to propagate the flag to the script that is being executed. I have updated the PR

sudiptob2 (approved):
LGTM

richscott (approved):

Issue #171, opened 2024-08-27
Execution panic on concurrent read/write access of event-types counter map


richscott:
In a recent CI execution for an unrelated PR of the YHS unit tests, there was a fatal panic due to `concurrent map read and map write`, with the last code execution at `internal/yunikorn/stream_test.go:85` - the line is:
```
return eventCounts[expectedKey1] == 2 && eventCounts[expectedKey2] == 1
```
`eventCounts` is a `model.EventTypeCounts`, which is an alias for `map[string]int`. 

Running `make test-go-unit` 100 times resulted in 5 failures, so I expect in a production setting there will be runtime panics (and container restarts) observed fairly quickly, before this bug gets fixed.

PR #173, opened 2024-08-27
Event-Type Counts fix - lock event-counts mutex and return a map copy, to avoid panics

richscott:
For the InMemoryEventRepository.Counts() func, lock the internal mutex, then construct a copy of the event-counts map and return the copy - to avoid Golang "concurrent map read/write" panics.

Fixes https://github.com/G-Research/yunikorn-history-server/issues/171

### DIFF
--- a/internal/database/repository/events.go
+++ b/internal/database/repository/events.go
@@ -31,7 +31,17 @@ func NewInMemoryEventRepository() *InMemoryEventRepository {
 }
 
 func (r *InMemoryEventRepository) Counts(ctx context.Context) (model.EventTypeCounts, error) {
-	return r.counts, nil
+	// We must lock and make a copy of the original map to avoid
+	// "concurrent map read and map write" panics, if the caller
+	// of this func reads from the returned result of this func.
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+	countsCopy := model.EventTypeCounts{}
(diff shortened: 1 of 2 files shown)


sudiptob2 (commented):
Nice catch! 🚀  Is it possible to implement a unit test for this case? We first saw this problem in `ci` but it was not deterministic. Is there a reliable way to test this function?

richscott:
> Nice catch! 🚀 Is it possible to implement a unit test for this case? We first saw this problem in `ci` but it was not deterministic. Is there a reliable way to test this function?

@sudiptob2 It's basically a race-condition situation, so there's no way (that I know of) to enforce a constant deterministic way to provoke the bug. However, I just added a unit-test (as usual, the unit-test is about 3x as long as the original fix 😅 ) that fires off 10 goroutines, each one calls EventRepository.Record(eventObject, EventType), then sleeps for a random few milliseconds, then does it again, for a few hundred times in each goroutine. Meanwhile, there is a `assert.Eventually()` test check that runs every 5 ms and just calls the EventRepository.Counts() function to get its map (it actually gets a copy, for safety) - in the `main` branch, when I `make test-go-unit` 100 times, it now fails approximately 17 times (a 17% failure rate on the unit-tests overall), on this branch, it never fails.

dave-gantenbein (approved):

PR #174, opened 2024-08-28
Implement queue event handler.

### LINKED ISSUE #147: Add event-handler logic for EventRecord_QUEUE type
Currently, the only YK EventRecord type that YHS handles is `EventRecord_APP`.  We need to implement logic to process `EventRecord_QUEUE` events, for change-types of ADD  and REMOVE, and possibly also SET. This logic should call the `UpsertQueues(...)` method to persist the new queue to YHS. Also, we should track the given `parent` and use that to find the `parent_id` to be used when inserting this new queue.

- [x] ADD change-event for QUEUE
- [x] REMOVE change-event for QUEUE
- [x] SET (optional?) change-event for QUEUE

sudiptob2:
Closes #147 

- [x] Implement QueueADD event handler


nikola-jokic:
Can we close this one?

PR #190, opened 2024-08-30
Implement GetQueue repo method

### LINKED ISSUE #183: Implement GetQueue() function in db repository layer

sudiptob2:
Closes #183

Returns a single `queue` with all its children embedded recursively in the `queue.Children` property.

**Note:**  
The following query is used to fetch the queue:
```sql
selectSQL := `SELECT * FROM queues WHERE partition = $1`
```
I am fetching all queues in the partition at once. This method when used in combination with `getChildrenFromMap`, will minimize the number of database calls while building `children` tree.


internal/database/repository/queue.go
```diff
@@ -190,6 +190,59 @@ func (s *PostgresRepository) GetQueuesPerPartition(
 	return queues, nil
 }
 
+// GetQueue GetQueuesPerPartition the queue with the given name and partition
+// child queues are nested in the queue.Children field
+func (s *PostgresRepository) GetQueue(ctx context.Context, partition, queueName string) (*model.PartitionQueueDAOInfo, error) {
+	selectSQL := `SELECT * FROM queues WHERE partition = $1`
```
richscott, line 196: I understand the drive to try to minimize query round-trips to the database, but I think this might actually be slower (maybe a lot slower) in production.  If the YHS database stores many months of past data, this query could end up fetching many thousands of rows, since we're getting every row in the `queues` table that belonged to the specified partition. We then run `rows.Scan()` on every row, build up a large temporary map just to get a list of all possible children for every queue (row entry), attach all the children from the map for that queue (based on its ID), and then return that.

I think it would be much more efficient to use that "parent and all descendants" query described in https://learnsql.com/blog/query-parent-child-tree/ -- see the "Example 4: List all Descendants" section.  It would give you the matching queue and all its descendants in a single query. The code would be something like:
```
WITH RECURSIVE generation AS (
    SELECT id,
         queue_name,
         /* other fields from queues table here.... */
         parent_id,
         0 AS generation_number
    FROM queues
    WHERE parent_id IS NULL
 
UNION ALL
 
    SELECT child.id,
         child.queue_name,
         child.parent_id,
         generation_number+1 AS generation_number
    FROM queues child
    JOIN generation g
      ON g.id = child.parent_id
 
)
 
SELECT  g.queue_name AS child_queue_name,
        g.generation_number,
        parent.queue_name AS parent_queue_name,
        /* other fields here */
FROM generation g
WHERE g.queue_name = $1 AND g.partition = $2  -- AND deleted_at IS NULL
LEFT JOIN queues parent
ON g.parent_id = parent.id
ORDER BY generation_number; 
```

I think this would be very efficient, both in the database, and also in terms of network traffic and how much processing this func would do with the returned data.

richscott:
I had a suggestion about the `GetQueue()` function. Also, I think the `Lint` CI check is complaining that `queue_int_test.go` needs to have `goimports -w` run on it.

internal/database/repository/queue.go
sudiptob2: Thanks 🙏  implemented with some modifications.

dave-gantenbein (approved):

internal/database/repository/queue_int_test.go
```diff
@@ -103,6 +104,59 @@ func TestGetQueuesPerPartition_Integration(t *testing.T) {
...
+		{
+			name:                     "Get Queue root.org.eng",
+			queueName:                "root.org.eng",
+			partition:                "default",
+			expectedChildrenQueueLen: 2,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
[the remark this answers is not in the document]
sudiptob2: Added 🔢

dave-gantenbein (approved):

dave-gantenbein (approved):

PR #199, opened 2024-09-10
Add CORS middleware

### LINKED ISSUE #196: Implement CORS for YHS webservice APIs
## Problem
When the YHS web client attempts to access the web service APIs, it encounters a `missing CORS "Access-Control-Allow-Origin" header` error. This issue arises because Cross-Origin Resource Sharing (CORS) is not configured in the **YHS web service backend**. Without CORS implementation, the browser automatically blocks requests made from different origins for security reasons, preventing the client from successfully interacting with the APIs.

## Possible Solution
Add a `CORS` handler in the web service. [rs/cors](https://github.com/rs/cors) could be useful.

sudiptob2:
closes #196

## Manual Testing
Also done following manual testing

### If origin is not allowed CORS
The browser rejects the response and gives a CORS error in the console.

image

### If origin is allowed in CORS
The browser accepts the response and successfully displays the webpage.

image

### DIFF
--- a/config/yunikorn-history-server/config.yml
+++ b/config/yunikorn-history-server/config.yml
@@ -17,6 +17,16 @@ db:
 yhs:
   port: 8989
   data_sync_interval: 5m
+  cors:
+    allowed_origins:
+      - "*"
+    allowed_methods:
+      - "GET"
+      - "POST"
+      - "PUT"
+      - "DELETE"
+    allowed_headers:
+      - "*"
 
 log:
   level: "INFO"
--- a/config/yunikorn-history-server/config.yml
+++ b/config/yunikorn-history-server/config.yml
@@ -22,9 +22,6 @@ yhs:
       - "*"
     allowed_methods:
       - "GET"
-      - "POST"
-      - "PUT"
-      - "DELETE"
     allowed_headers:
       - "*"
 
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -7,7 +7,7 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
-
+	"github.com/rs/cors"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -34,6 +34,11 @@ func TestNew(t *testing.T) {
 					Port:             8080,
 					AssetsDir:        "assets",
 					DataSyncInterval: 5 * time.Minute,
+					CORSConfig: cors.Options{
+						AllowedOrigins: []string{"*"},
+						AllowedMethods: []string{"GET", "POST", "PUT", "DELETE"},
+						AllowedHeaders: []string{"*"},
+					},
 				},
 				YunikornConfig: YunikornConfig{
 					Host:   "localhost",
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -36,7 +36,7 @@ func TestNew(t *testing.T) {
 					DataSyncInterval: 5 * time.Minute,
 					CORSConfig: cors.Options{
 						AllowedOrigins: []string{"*"},
-						AllowedMethods: []string{"GET", "POST", "PUT", "DELETE"},
+						AllowedMethods: []string{"GET"},
 						AllowedHeaders: []string{"*"},
 					},
 				},
--- a/internal/webservice/webservice.go
+++ b/internal/webservice/webservice.go
@@ -48,7 +48,7 @@ func (ws *WebService) Start(ctx context.Context) error {
 	logger = logger.With("component", "webservice")
 	ctx = log.ToContext(ctx, logger)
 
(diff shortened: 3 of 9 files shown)


cmd/yunikorn-history-server/commands/root.go
```diff
@@ -77,7 +77,7 @@ func Run(ctx context.Context, cfg *config.Config) error {
 
 	healthService := health.New(info.Version, health.NewYunikornComponent(client), health.NewPostgresComponent(pool))
 
-	ws := webservice.NewWebService(&cfg.YHSConfig, mainRepository, eventRepository, healthService)
+	ws := webservice.NewWebService(&cfg.YHSConfig, &cfg.CORSConfig, mainRepository, eventRepository, healthService)
```
nikola-jokic, line 80: Since cors is associated with server config, I think it might be better to have it in the `YHSConfig` struct

internal/webservice/webservice.go
```diff
@@ -47,6 +54,8 @@ func (ws *WebService) Start(ctx context.Context) error {
 
 	ws.initRoutes(ctx)
 
+	ws.setupMiddlewares()
```
nikola-jokic, line 57: Wouldn't it be better if middlewares are created within `initRoutes`? That way, initRoutes can install all handlers and middlewares that it needs. After the `initRoutes` call, you know that routes are created and appropriate middlewares are installed. This doesn't need to be a two-step process :relaxed:

internal/webservice/webservice.go
```diff
@@ -19,10 +21,13 @@ type WebService struct {
 	eventRepository repository2.EventRepository
 	healthService   health.Interface
 	assetsDir       string
+	corsConfig      cors.Options
+	router          *httprouter.Router
```
nikola-jokic, line 25: Might be better not to keep the router in a struct since the only use of it is to build routes and install it on `ws.server.Handler`

internal/webservice/webservice.go
sudiptob2: Thanks, I was thinking about and could not decide which one would be better. Will update

config/yunikorn-history-server/config.yml
```diff
@@ -21,3 +21,14 @@ yhs:
 log:
   level: "INFO"
   json_format: false
+
+cors:
```
dejanzele, line 25: As per @nikola-jokic's comment, this should live under the `yhs` block.

internal/webservice/webservice.go
dejanzele: the method could also be rename `initRoutes` -> `init`

config/yunikorn-history-server/config.yml
sudiptob2: Updated. 
I think we need to change the implementation of `processEnvVar`.  We support configuration through both the config files and ENV vars.  According to the current implementation, there is no way to define nested env variables of **deeper than a single level** using ENV vars.

For example, if we try to define the following cors configuration:

`YHS_YHS_CORS_ALLOWED_ORIGIN` `(prefix_nest1_next2_iam_nest_3)` will be converted to `yhs.cors_allowed_origin` but from the configuration file it is read as `yhs.cors.allowed_origin`

---
**Relevant code:**
https://github.com/sudiptob2/yunikorn-history-server/blob/c912b75a837050853da9917306dc5509636658ce/internal/config/config.go#L195-L211
```go
	if err := k.Load(env.Provider("YHS_", ".", processEnvVar), nil); err != nil {
		return nil, fmt.Errorf("error loading environment variables: %v", err)
	}


	return k, nil
}


// Removes the prefix "YHS_" and replaces the first "_" with "."
// YHS_PARENT1_CHILD1_NAME will be converted into "parent1.child1_name"
func processEnvVar(s string) string {
	s = strings.TrimPrefix(s, "YHS_")
	firstIndex := strings.Index(s, "_")
	if firstIndex > -1 {
		s = s[:firstIndex] + "." + s[firstIndex+1:]
	}
	return strings.ToLower(s)
}
```

nikola-jokic (dismissed):
LGTM

internal/config/config_test.go
```diff
@@ -34,6 +34,11 @@ func TestNew(t *testing.T) {
 					Port:             8080,
 					AssetsDir:        "assets",
 					DataSyncInterval: 5 * time.Minute,
+					CORSConfig: cors.Options{
+						AllowedOrigins: []string{"*"},
+						AllowedMethods: []string{"GET", "POST", "PUT", "DELETE"},
```
dejanzele, line 39: Do we have DELETE and PUT endpoints?

sudiptob2: Yeah, I think we can remove those. We only do GET.

nikola-jokic (approved):

PR #203, opened 2024-09-11
Add codegen job to check that generated code is up-to-date

### LINKED ISSUE #95: Add a CI job for checking generated code is up-to-date

nikola-jokic:
Fixes #95

### DIFF
--- a/.github/actions/setup-go/action.yml
+++ b/.github/actions/setup-go/action.yml
@@ -14,7 +14,7 @@ runs:
     - id: setup-go
       uses: actions/setup-go@v5
       with:
-        go-version: "1.21"
+        go-version-file: "go.mod"
         check-latest: true
         cache: false
     - id: cache-info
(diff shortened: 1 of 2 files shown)


dave-gantenbein (approved):
lgtm

pavlovic-ivan:
Hi @nikola-jokic . Thx for the PR. Can you maybe list pros and cons of using `go-version-file` vs `go-version`?

nikola-jokic:
Of course! The `go.mod` file specifies the go version used when building, testing, etc. The CI only does that automatically, so ideally, we need to set up the specified environment. We should rely on `go.mod` wherever we can, since it is the source of truth, and specifies the toolchain required (especially from go 1.21 when this change was made).
Just as an example, we used setup go with version `1.21`, while our `go.mod` specified `1.22.4`. And we should avoid manual updates to our CI when we can reference it.

pavlovic-ivan:
Sounds reasonable. The only reason why version was specified explicitly in the workflow, was because of possible drift between Github adopting the new version and official release of Golang. If GIthub is keeping up nicely with releases, i don't mind having this change

PR #206, opened 2024-09-12
Use 'yhs' as a default database instead of 'postgres'

### LINKED ISSUE #84: Use 'yhs' as default Postgres database instance name for YHS services
It's bad practice - very, very bad practice - to use the `postgres` db instance for a running service, unless you explicitly need to instantiate new db instances as part of run-time operations. The YHS service should use `yhs` or any other name that is not in the set `(postgres, template0, template1)` for its database.

nikola-jokic:
Fixes #84

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,9 +129,19 @@ define yq_get_db
     $(shell $(call yq_get, .db.$(1)))
 endef
 
+DB_USER ?= $(strip $(call yq_get_db,user))
+DB_PASSWORD ?= $(strip $(call yq_get_db,password))
+DB_HOST ?= $(strip $(call yq_get_db,host))
+DB_PORT ?= $(strip $(call yq_get_db,port))
+DB_NAME ?= $(strip $(call yq_get_db,dbname))
+
 define database_url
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -129,10 +129,20 @@ define yq_get_db
...
+
 define database_url
-	postgres://$(strip $(call yq_get_db,user)):$(strip $(call yq_get_db,password))@$(strip $(call yq_get_db,host)):$(strip $(call yq_get_db,port))/$(strip $(call yq_get_db,dbname))?sslmode=disable
+	postgres://$(DB_USER):$(DB_PASSWORD)@$(DB_HOST):$(DB_PORT)/$(DB_NAME)?sslmode=disable
 endef
 
+.PHONY: create-postgres-db
+create-postgres-db: ## create database specified in the config file.
+	PGPASSWORD=$(DB_PASSWORD) psql -h $(DB_HOST) -p $(DB_PORT) -U $(DB_USER) -c "CREATE DATABASE $(DB_NAME);"
+
```
sudiptob2, line 145: Will this work if `psql` is not installed in the environment? I think we should add the `psql` installation under the `install-dependencies` command.

sudiptob2 (changes requested):

Makefile
nikola-jokic: Makes sense, I moved away from psql to avoid installing it and used `values.yaml` file when setting up postgres to create db right away

sudiptob2 (approved):

dejanzele (approved):

PR #210, opened 2024-09-12
Flatten out the config to allow similar yaml and env fetching

### LINKED ISSUE #204: Inconsistent Parsing Between YAML Config and Environment Variables for Nested Fields
We support configuration through both config files and environment variables (ENV vars). In the current implementation, it is not possible to define nested environment variables **beyond a single level** using ENV vars. Let's clarify this with an example.

Consider the following CORS configuration defined in YAML:

```yaml
yhs:
  port: 8989
  data_sync_interval: 20s
  cors:
    allowed_origins:
      - "*"
    allowed_methods:
      - "GET"
    allowed_headers:
      - "*"
```
If we want to define this using env vars, we would have to do something like the following:

```
YHS_YHS_CORS_ALLOWED_ORIGIN=*
...
```
`YHS_YHS_CORS_ALLOWED_ORIGIN` would be converted to yhs.cors_allowed_origin by the `processEnvVar()` function, but in the configuration file, it is read as `yhs.cors.allowed_origin`.

**As a result, the parsing of YAML and ENV vars is inconsistent.**

---
**Relevant code:**
https://github.com/G-Research/yunikorn-history-server/blob/ce8153c8910a898216d16d9fe3c97fbbf59bfe7d/internal/config/config.go#L173-L201

nikola-jokic:
YAML format is structured, allowing multiple levels within an object. However, envs are flat (only one level).

This change ensures that yaml is flattened, so fetching environment variables is the same for both formats. It still allows multiple levels, but both of these will be accepted:
```yaml
# This change allows hierarchical structure
yhs:
  port: 8080
# And the flat structure
yhs_port: 8080
```

Fixes #204

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -120,11 +120,11 @@ func New(path string) (*Config, error) {
 		return nil, err
 	}
 
-	assetsDir := k.String("yhs.assets_dir")
+	assetsDir := k.String("yhs_assets_dir")
 	if assetsDir == "" {
 		assetsDir = "assets"
 	}
-	dataSyncInterval := k.Duration("yhs.data_sync_interval")
+	dataSyncInterval := k.Duration("yhs_data_sync_interval")
 	if dataSyncInterval == 0 {
 		dataSyncInterval = 5 * time.Minute
 	}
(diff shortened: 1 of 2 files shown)


sudiptob2 (approved):

internal/config/config.go
```diff
@@ -202,10 +204,5 @@ func loadConfig(cfgFile string) (*koanf.Koanf, error) {
 // Removes the prefix "YHS_" and replaces the first "_" with "."
 // YHS_PARENT1_CHILD1_NAME will be converted into "parent1.child1_name"
```
sudiptob2, line 205: This docstring should change

sudiptob2 (dismissed):

internal/config/config.go
nikola-jokic: good catch!

sudiptob2 (approved):
LGTM

PR #211, opened 2024-09-13
Implement SPA handler to return ui on root path

### LINKED ISSUE #59: Add API endpoint to deliver YHS web-ui JS bundle(s)
The YHS front-end web application will need some custom bundles of Javascript (+ any other resources?) to be delivered from the YHS api service. It's expected that the bundle(s) will be static files, and that they will be constructed as part of the regular YHS build and release process via Github CI processing - the bundles will be built and delivered as part of the YHS Docker iamge. The actual endpoint is yet to be decided, but a simple static file-server route handler can be constructed in the meantime.

nikola-jokic:
SPA handler returns index.html if the file requested is not present, so it can leave routing to the UI framework.

Fixes #59

### DIFF
--- a/internal/webservice/routes.go
+++ b/internal/webservice/routes.go
@@ -3,6 +3,8 @@ package webservice
 import (
 	"context"
 	"net/http"
+	"os"
+	"path/filepath"
 
 	"github.com/google/uuid"
 	"github.com/julienschmidt/httprouter"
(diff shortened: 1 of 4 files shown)


sudiptob2 (commented):
🚀 . Is it possible to add any **test** for this change?

sudiptob2 (approved):
LGTM

PR #217, opened 2024-09-16
Modify response for get queues by partition so the queue tree is returned

### LINKED ISSUE #215: GetQueuesPerPartition returns a List

### LINKED ISSUE #216: Response from `/ws/v1/partition/:partition_name/queues` does not match the same format used by Yunikorn
Previously, the list of objects was returned by YHS, while Yunikorn returns a root object along with its children.
Call to [GetQueuesPerPartition](https://github.com/G-Research/yunikorn-history-server/blob/3aa2a22f4cd8077c05cd2c2d91d7891c6fc8907d/internal/database/repository/repository.go#L33) returned list of objects from `model` package, that was later [serialized](https://github.com/G-Research/yunikorn-history-server/blob/3aa2a22f4cd8077c05cd2c2d91d7891c6fc8907d/internal/webservice/routes.go#L110) in response.

The correct response should be a single root object from Yunikorn's `dao` package, that contains child queues.

nikola-jokic:
This change modifies the repository as well as the API response in following ways:
- `repository.GetQueuesPerPartition` now returns all queues associated with partition. It is left to the caller to build a hierarchical view.
- `webservice.getQueuesPerPartition` unpacks the `model` and builds up the tree from `dao` objects. It will not include internal fields used by `model`.

Fixes #216
Fixes #215


internal/webservice/routes.go
```diff
@@ -107,7 +110,50 @@ func (ws *WebService) getQueuesPerPartition(w http.ResponseWriter, r *http.Reque
...
+	for _, queue := range queues {
+		queueMap[queue.Id] = &queue.PartitionQueueDAOInfo
+	}
+
+	var rootID string
+	for _, queue := range queues {
+		if !queue.ParentId.Valid {
+			rootID = queue.Id
+			continue
+		}
```
sudiptob2, line 139: What if we have multiple root queues marked as deleted? In `yunikorn`, it's not possible to have more than one active root queue at the same time per partition. However, since we are displaying historical data, we may need to show queues that have been removed as well.

Should the front end adapt to handle a list of root queues instead? Restricting the system to return only one root queue may not be suitable for our use case.

nikola-jokic: Great point!

internal/webservice/routes.go
```diff
@@ -239,11 +241,55 @@ func (ws *WebService) getQueuesPerPartition(req *restful.Request, resp *restful.
...
 	}
-	daoQueues := make([]*dao.PartitionQueueDAOInfo, 0, len(queues))
+	root, err := buildPartitionQueueTrees(ctx, queues)
+	if err != nil {
+		errorResponse(req, resp, err)
+	}
+	jsonResponse(resp, root)
+}
+
+func buildPartitionQueueTrees(ctx context.Context, queues []*model.PartitionQueueDAOInfo) ([]*dao.PartitionQueueDAOInfo, error) {
```
sudiptob2, line 251: Just a quick question: Wouldn't it be better to return `[]*model.PartitionQueueDAOInfo` since it includes additional fields like created_at and deleted_at, which could be useful for the frontend when implementing designs to highlight active or deleted queues?

nikola-jokic: Right, then we would take the hybrid approach, I'll work on that, but it does make sense to me.

sudiptob2: Yeah, today or tomorrow we have to support additional fields for our use case.

PR #221, opened 2024-09-17
Rework Queue sync logic

### LINKED ISSUE #186: Rework queue sync logic to use new queues DB methods
related to #143

For the "sync" processing logic, we need to migrate the code from using the old, deprecated `UpsertQueues()` func to the new CRUD-style functions. In `internal/yunikorn/sync.go`, in `upsertPartitionQueues()`, the code should now:
- query the db for all non-deleted queues it has stored
   - for each non-deleted queue in db,  if the YK response does not list that queue, call `DeleteQueue()` with that queue as argument
- for each queue in YK response, check if queue is in YHS db by calling `GetQueue(ctx, partition, queueName)`
    - if not, call `AddQueue(ctx, queues)`
    - if it is in the db, call `UpdateQueue(ctx, queueID, queueDAO)` with (possibly new) field values

sudiptob2:
closes #186


internal/database/repository/queue.go
```diff
@@ -188,6 +188,9 @@ func (s *PostgresRepository) UpdateQueue(ctx context.Context, queue *dao.Partiti
         allocating_accepted_apps = @allocating_accepted_apps
     WHERE queue_name = @queue_name AND partition = @partition AND deleted_at IS NULL
 `
+	if queue.Partition == "" {
```
nikola-jokic, line 191: 2 questions:
1. Can you please move the if above the query? It is easier when I read this method to check the invariants, then see the query knowing that all invariants hold
2. In the future, do we implement proper validations on each object like `ValidDAOParititionQueueDAOInfo`? This is the open question and definitely not for this PR, but I just wanted to mention it

internal/yunikorn/sync.go
```diff
@@ -94,55 +95,117 @@ func (s *Service) upsertPartitions(ctx context.Context) ([]*dao.PartitionInfo, e
...
-	var errs []error
-	var queues []*dao.PartitionQueueDAOInfo
+	addQueues := func(queues []*dao.PartitionQueueDAOInfo) {
+		mu.Lock()
+		defer mu.Unlock()
+		allQueues = append(allQueues, queues...)
+	}
+
+	addJob := func(partition *dao.PartitionInfo) {
+		wg.Add(1)
```
nikola-jokic, line 120: wg.Add should never be within a function. This creates a hard-to-find race condition where, if this goroutine is not yet started, and others signal with wg.Done, the counter can reach 0 without having this function executed.

internal/yunikorn/sync.go
```diff
@@ -94,55 +95,117 @@ func (s *Service) upsertPartitions(ctx context.Context) ([]*dao.PartitionInfo, e
...
-		} else {
-			queues = append(queues, queue)
+			logger.Errorf("could not add sync queues job for partition %s: %v", partition.Name, err)
 		}
 	}
 
-	for _, p := range partitions {
-		go processPartition(p)
+	for _, partition := range partitions {
+		addJob(partition)
```
nikola-jokic, line 141: I think we are complicating too much here. I'm not sure what do we gain from pushing sync of queues to the background. The sync partitions is pushed to the background. That is a single unit of work. Then, correctly synchronized partition is the one that synchronized the queue. Since we are waiting on the workqueue, it is effectively the same as waiting on the goroutine. So why not use the language primitive?

The idea of the workqueue is (from my point of view) to do something in the background. Usually, best-effort cleanups and such should go there. But since we care about the result, we should simply leverage goroutines to make our life easier. The way we used workqueue here is the same way we would use the goroutine:

```go
errs := make(chan error, len(partitions))
partitionQueues := make(chan []*dao.PartitionQueueDAOInfo, len(partitions))
wg.Add(len(partitions))
for _, partition := range partitions {
  go func() {
    defer wg.Done()
    queues, err := syncPartitionQueues(partition)
    if err != nil {
      errs <- err
      return
    }
    partitionQueues <- queues
  }()
}
wg.Wait()
close(errs)
close(partitionQueues)
var syncErrors []error
for err := range errs {
  syncErrors = append(syncErrors, err)
}
if len(syncErrors) > 0 {
  return errors.Join(syncErrors...)
}
var queues []*dao.PartitionQueueDAOInfo
for qs := range partitionQueues {
  queues = append(queues, qs...)
}
return queues, nil
```

internal/yunikorn/sync.go
```diff
@@ -94,55 +95,117 @@ func (s *Service) upsertPartitions(ctx context.Context) ([]*dao.PartitionInfo, e
...
+
+func (s *Service) syncPartitionQueues(ctx context.Context, partition *dao.PartitionInfo) ([]*dao.PartitionQueueDAOInfo, error) {
+	// Fetch partition queues from the YuniKorn API
+	queue, err := s.client.GetPartitionQueues(ctx, partition.Name)
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve queues from YuniKorn API: %w", err)
+	}
+
+	// Attempt to update the queue; if it fails, try adding it instead
+	if err := s.repo.UpdateQueue(ctx, queue); err != nil {
```
nikola-jokic, line 161: I'm not sure why we don't simply use upsert, but let's leave it for later

nikola-jokic (commented):
Left few comments. lmkwyt :relaxed:

internal/database/repository/queue.go
sudiptob2: Seems to be a good idea.

internal/yunikorn/sync.go
sudiptob2: we used to use `upsertQueue` method. But the idea of https://github.com/G-Research/yunikorn-history-server/issues/143 was to separate out CRUD and move away from upsert. I think we have parent-child relationship between queues which is very difficult to update using the upsert.

nikola-jokic (approved):
lgtm!

PR #225, opened 2024-09-19
Solve `AddQueue` events and sync partitions on `AddQueue` event.

nikola-jokic:
When `add queue` event is raised, the information within the payload is incomplete to store the history. We get notified about the queue name, but we don't know what partition it belongs to, or if a new partition is being created.

Current version `1.5.2` does not include events based on new partitions, but this implementation should take care of it.

The way it is implemented is:
1. `accumulator` accumulates events
2. When events are idle for `idleInterval`, it calls upsert methods to sync partitions and queues.

Since this operation is not frequent, and since we don't have enough information currently within the event, both add queues and add partitions are going to be handled correctly, and without putting too much pressure on Yunikorn.

This PR is related to #147


internal/yunikorn/event_handler.go
```diff
@@ -234,3 +235,47 @@ func (s *Service) handleAppRemoveEvent(ctx context.Context, ev *si.EventRecord)
...
+func (s *Service) handleQueueAddEvent(ctx context.Context) {
+	logger := log.FromContext(ctx)
+
+	partitions, err := s.upsertPartitions(ctx)
+	if err != nil {
+		logger.Errorf("could not get partitions: %v", err)
+		return
+	}
+
+	if _, err := s.upsertPartitionQueues(ctx, partitions); err != nil {
```
sudiptob2, line 277: We will be removing `upsertPartitionQueues` as per https://github.com/G-Research/yunikorn-history-server/issues/143

A work in progress in https://github.com/G-Research/yunikorn-history-server/pull/221 to use the newly added CRUD.

internal/yunikorn/event_handler.go
```diff
@@ -234,3 +235,47 @@ func (s *Service) handleAppRemoveEvent(ctx context.Context, ev *si.EventRecord)
...
+	case si.EventRecord_SET:
+		// Ignored for now
+	case si.EventRecord_REMOVE:
+		// Ignored for now
+	case si.EventRecord_NONE:
+	default:
+		// should be warning
+		logger.Warnf("unknown event EventChangeType for an Event of type QUEUE: %v", ev.GetEventChangeType())
+	}
+}
```
sudiptob2, line 254: According to https://github.com/G-Research/yunikorn-history-server/issues/186, the sync logic will manage changes in the queues, including Add, Delete, and Update operations. 

Can we say that the same accumulator can be used for all the above cases?

nikola-jokic: Yes, but each of these cases would require special logic. What I wanted to do is prepare the event store for the Add event, and then handle deletion events in a separate PR. This PR is pretty big right now, so I would ideally want to break it down.
Breaking it down would make it easier to review and explain changes.

sudiptob2 (approved):

sudiptob2 (approved):

PR #243, opened 2024-09-27
Add partition delete logic in `sync`

### LINKED ISSUE #239: Implement partition delete
- Add a `deleted_at` field to the `partition` table.
- Implement the `deletePartition()` method in the repository.
- Update the sync logic:
  - Identify which partitions have been removed by comparing the current partitions in the database with those retrieved from the API.
  - Mark the removed partitions by setting the `deleted_at` field.

sudiptob2:
closes #239


internal/database/repository/partitions.go
```diff
@@ -54,19 +57,18 @@ func (s *PostgresRepository) UpsertPartitions(ctx context.Context, partitions []
 	return nil
 }
 
-func (s *PostgresRepository) GetAllPartitions(ctx context.Context) ([]*dao.PartitionInfo, error) {
+func (s *PostgresRepository) GetAllPartitions(ctx context.Context) ([]*model.PartitionInfo, error) {
```
dave-gantenbein, line 60: Should we be preparing for pagination parameters here? It seems unlikely we'd ever want to return everything to the front end all at once...

sudiptob2: Thanks, we also have several other APIs that need to be updated with pagination. This task is tracked in https://github.com/G-Research/yunikorn-history-server/issues/245

internal/yunikorn/sync.go
```diff
@@ -84,16 +84,49 @@ func (s *Service) upsertPartitions(ctx context.Context) ([]*dao.PartitionInfo, e
...
-		logger.Infow("upserting partitions", "count", len(partitions))
-		return s.repo.UpsertPartitions(ctx, partitions)
-	}, workqueue.WithJobName("upsert_partitions"))
+		logger.Infow("syncing partitions", "count", len(partitions))
+		err := s.repo.UpsertPartitions(ctx, partitions)
+		if err != nil {
+			return fmt.Errorf("could not upsert partitions: %w", err)
+		}
+		// Delete partitions that are not present in the API response
+		deleteCandidates, err := s.findPartitionDeleteCandidates(ctx, partitions)
```
nikola-jokic, line 93: since workqueue has retries, you should probably build up `deleteCandidates` before pushing the task on it. If we ever retry, there is no need to re-calculate them

internal/database/repository/partitions.go
```diff
@@ -88,3 +91,19 @@ func (s *PostgresRepository) GetAllPartitions(ctx context.Context) ([]*dao.Parti
...
 }
+
+func (s *PostgresRepository) DeletePartitions(ctx context.Context, partitions []*model.PartitionInfo) error {
+	partitionNames := make([]string, len(partitions))
+	for i, p := range partitions {
+		partitionNames[i] = p.Name
+	}
+
+	deletedAt := time.Now().Unix()
+	query := `UPDATE partitions SET deleted_at = $1 WHERE name = ANY($2)`
```
nikola-jokic, line 102: You should probably target these partitions by ID. Imagine that we had a partition named "example". Then we removed it. Then we re-installed it. Then we remove it again. Currently, in the database, we have multiple partition records with the name "example", but only one of them should have `deleted_at` set to now.

internal/database/repository/partitions.go
nikola-jokic, line 60: Here, you should probably get only active partitions. We don't need to get deleted partitions to operate on them.
We should also change the name, so we know that we are fetching active partitions.

nikola-jokic (changes requested):

internal/database/repository/partitions.go
sudiptob2: Hi, Thanks for the awesome reviews 🚀 
I found that, `GetAllPartitions` is also being used by the `webservice` [here](https://github.com/G-Research/yunikorn-history-server/blob/04d00cef9427529ca5e84809be925f076ead5058/internal/webservice/routes.go#L227). I think frontend might need the deleted partition data as well to show some graphical design on them. Therefore, I updated the logic in `findPartitionDeleteCandidates()`

internal/yunikorn/sync.go
```diff
@@ -74,26 +74,59 @@ func (s *Service) sync(ctx context.Context) error {
...
 
+func (s *Service) findPartitionDeleteCandidates(ctx context.Context, apiPartitions []*dao.PartitionInfo) ([]*model.PartitionInfo, error) {
+
+	apiPartitionMap := make(map[string]*dao.PartitionInfo)
+	for _, p := range apiPartitions {
+		apiPartitionMap[p.Name] = p
+	}
+
+	// Fetch partitions from the database
+	dbPartitions, err := s.repo.GetAllPartitions(ctx)
```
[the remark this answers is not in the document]
sudiptob2: Yeah true, I found that, GetAllPartitions is also being used by the web service [here](https://github.com/G-Research/yunikorn-history-server/blob/04d00cef9427529ca5e84809be925f076ead5058/internal/webservice/routes.go#L227). Should we embed the `is null` logic in `GetAllPartitions()` as this method is not used in internal call **only**.

I think we have the following two approaches here,
1. Implement a different `GetAllActivePartition()` for internal call.
2. once we add support for pagination and filtering we will be able to filter the `null only` partitions in our internal calls when needed. 

Let me know your thoughts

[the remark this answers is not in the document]
sudiptob2: Implemented seperate `GetActivePartitions()`

nikola-jokic (approved):
We can push the work of finding partitions to the database to do something like:
```sql
DELETE FROM partitions where id IN (
  select id from partitions where deleted_at IS NULL AND NOT(id = ANY($1))
)
```
But I will leave it up to you if you want to do it now or we can proceed with this for now

sudiptob2:
> We can push the work of finding partitions to the database to do something like:
Hmm.. seems cleaner, we can reduce code in this way 👍

internal/database/repository/partitions.go
```diff
@@ -88,3 +126,25 @@ func (s *PostgresRepository) GetAllPartitions(ctx context.Context) ([]*dao.Parti
 	}
 	return partitions, nil
 }
+
+func (s *PostgresRepository) DeletePartitions(ctx context.Context, partitions []*dao.PartitionInfo) error {
```
nikola-jokic, line 130: Rename to `DeleteActivePartitionsNotInPartitions` or something similar. The name is misleading :relaxed:

nikola-jokic (changes requested):

internal/database/repository/partitions.go
sudiptob2: Updated the name, but still not sure about the wording 😐. Let me know how it sounds.

nikola-jokic (approved):

PR #251, opened 2024-09-30
replace uuid with ulid

### LINKED ISSUE #242: Use ulid instead of uuid everywhere

sudiptob2:
closes #242


nikola-jokic (changes requested):
We might want to change the db model. Default is UUID which should be `text`
https://github.com/G-Research/yunikorn-history-server/blob/d848ae82e644b770cbc7b4a9204e5be816f27c9f/migrations/20240628155246_init-tables.up.sql#L21

nikola-jokic (approved):
lgtm

PR #263, opened 2024-10-07
implement pagination and filters to nodes endpoint

### LINKED ISSUE #247: Implement paginated get APIs for `Node`
We have the following two `repository` methods to serve `node` related data to the front end. We need to update these methods and add pagination and filter.

1. [GetNodesPerPartition()](https://github.com/G-Research/yunikorn-history-server/blob/04d00cef9427529ca5e84809be925f076ead5058/internal/database/repository/node.go#L79)

sudiptob2:
closes #247

### DIFF
--- a/internal/webservice/routes.go
+++ b/internal/webservice/routes.go
@@ -348,10 +348,25 @@ func (ws *WebService) getAppsPerPartitionPerQueue(req *restful.Request, resp *re
 	jsonResponse(resp, daoApps)
 }
 
+// getNodesPerPartition returns all nodes for a given partition and queue.
+// Results are ordered by creation time in descending order.
+// Following query params are supported:
+// - nodeId: filter by nodeId
+// - hostName: filter by hostName
+// - rackName: filter by rackName
+// - schedulable: boolean
+// - isReserved: boolean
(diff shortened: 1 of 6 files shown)


internal/webservice/routes.go
```diff
@@ -348,10 +348,25 @@ func (ws *WebService) getAppsPerPartitionPerQueue(req *restful.Request, resp *re
 	jsonResponse(resp, daoApps)
 }
 
+// getNodesPerPartition returns all nodes for a given partition and queue.
+// Results are ordered by creation time in descending order.
+// Following query params are supported:
+// - nodeId: filter by nodeId
```
richscott, line 354: As you do with `schedulable` and `isReserved` down below, maybe note the parameter type (e.g. string, int, etc) on these other parameters, to be consistent?

richscott (commented):
Just one doc suggestion, but I'll still approve if you don't want to change that item.

internal/webservice/routes.go
sudiptob2: Sure, updated. btw these comments are temporarily added aded for the front until we come up with some sort of documentation.

richscott (approved):

PR #267, opened 2024-10-09
implement pagination and filters to nodeutilizations endpoint

### LINKED ISSUE #248: Implement paginated get APIs for  `Node utilization`
We have the following two `repository` methods to serve `node_utilization` related data to the front end. We need to update these methods and add pagination and filter.

1. [GetNodeUtilizations()](https://github.com/G-Research/yunikorn-history-server/blob/04d00cef9427529ca5e84809be925f076ead5058/internal/database/repository/node.go#L79)

sudiptob2:
closes #248

### DIFF
--- a/internal/database/repository/node.go
+++ b/internal/database/repository/node.go
@@ -21,14 +21,14 @@ type NodeFilters struct {
 	Limit       *int
 }
 
-type NodeUtilizationFilters struct {
+type NodeUtilFilters struct {
 	ClusterID *string
 	Partition *string
 	Offset    *int
 	Limit     *int
 }
 
-func applyNodeUtilizationFilters(builder *sql.Builder, filters NodeUtilizationFilters) {
+func applyNodeUtilFilters(builder *sql.Builder, filters NodeUtilFilters) {
 	if filters.ClusterID != nil {
 		builder.Conditionp("cluster_id", "=", *filters.ClusterID)
 	}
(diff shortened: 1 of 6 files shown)


nikola-jokic (approved):
lgtm

internal/database/repository/node.go
```diff
@@ -7,8 +7,27 @@ import (
 	"github.com/apache/yunikorn-core/pkg/webservice/dao"
 	"github.com/jackc/pgx/v5"
 	"github.com/oklog/ulid/v2"
+
+	"github.com/G-Research/yunikorn-history-server/internal/database/sql"
 )
 
+type NodeUtilizationFilters struct {
```
richscott, line 14: What would you think about changing all occurrences of `NodeUtilizationXXX` to just `NodeUtilXXX`? I notice that the YK core uses the shorter style, like in `dao.PartitionNodesUtilDAOInfo`?   Besides being closer to the YK core style, it would be much less typing!

richscott (commented):
Just one naming suggestion for now...

internal/database/repository/node.go
sudiptob2: Updated. I left some cz for those utilization sounded better. Let me know what do you think

richscott (dismissed):
Perfect change, I agree about leaving the func names longer.

richscott (approved):

PR #268, opened 2024-10-10
implement pagination and filter in partition APIs

### LINKED ISSUE #249: Implement paginated get APIs for  `Partition`

sudiptob2:
closes #249

### DIFF
--- a/internal/database/repository/mock_repository.go
+++ b/internal/database/repository/mock_repository.go
@@ -55,6 +55,20 @@ func (mr *MockRepositoryMockRecorder) AddQueues(arg0, arg1, arg2 any) *gomock.Ca
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "AddQueues", reflect.TypeOf((*MockRepository)(nil).AddQueues), arg0, arg1, arg2)
 }
 
+// DeleteInactivePartitions mocks base method.
+func (m *MockRepository) DeleteInactivePartitions(arg0 context.Context, arg1 []*dao.PartitionInfo) error {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "DeleteInactivePartitions", arg0, arg1)
+	ret0, _ := ret[0].(error)
+	return ret0
+}
+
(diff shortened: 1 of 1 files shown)


internal/webservice/helpers.go
```diff
@@ -95,6 +146,16 @@ func parseHistoryFilters(r *http.Request) (*repository.HistoryFilters, error) {
 	return &filters, nil
 }
 
+func getClusterIDQueryParam(r *http.Request) string {
```
nikola-jokic, line 149: Since we want to set nil if value is not present, why don't we return `*string`. If the value is empty, return nil. otherwise, return pointer to the query string. For integer types, if value is empty, return nil. Otherwise, try to parse it. Then return the value of the parsed string or error

internal/webservice/helpers.go
```diff
@@ -11,16 +11,67 @@ import (
...
+	}
+	lastStateTransitionTimeEnd, err := getLastStateTransitionTimeEndQueryParam(r)
+	if err != nil {
+		return nil, err
+	}
+	if lastStateTransitionTimeEnd != nil {
+		filters.LastStateTransitionTimeEnd = lastStateTransitionTimeEnd
+	}
+	name := getNameQueryParam(r)
+	if name != "" {
```
nikola-jokic, line 46: Why don't we just instantiate the filter with fields like the following:
```go
filters.Name = r.URL.Query().Get("name")
```
Since we are not re-using constants, I don't think they add anything to us here.

nikola-jokic (changes requested):
Minor changes but overall, looks good

internal/webservice/helpers.go
sudiptob2: Updated

internal/webservice/helpers.go
sudiptob2: Actually, we need to set `nil` in case of empty string `""`. Instead, if we set `filters.Name` with an empty string (when no query param is passed) that will cause querybuilder to give incorrect result.

nikola-jokic (dismissed):
lgtm

nikola-jokic (approved):

PR #271, opened 2024-10-11
Update queue history based on the stream

### LINKED ISSUE #147: Add event-handler logic for EventRecord_QUEUE type

nikola-jokic:
This PR is based on #265

Fixes #147


internal/model/application.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"github.com/G-Research/yunikorn-core/pkg/webservice/dao"
+)
+
+type Application struct {
+	ModelMetadata          `json:",inline"`
+	dao.ApplicationDAOInfo `json:",inline"`
+}
+
+func (app *Application) MergeFrom(appInfo *dao.ApplicationDAOInfo) {
```
dave-gantenbein, line 12: Can we unit test any of this?

internal/yunikorn/stream.go
```diff
@@ -73,7 +64,18 @@ func (s *Service) processStreamResponse(ctx context.Context, response []byte) er
 		"change_detail", eventRecord.GetEventChangeDetail(),
 		"reference_id", eventRecord.GetReferenceID(),
 		"resource", eventRecord.GetResource(),
+		"state", eventRecord.GetState(),
 	)
 
+	// TODO: This is Okayish for small number of events, but for large number of events this will be a bottleneck
```
dave-gantenbein, line 70: Can you add a ticket instead of a TODO?

internal/yunikorn/sync.go
dave-gantenbein, line 1: Should there be testing of all this?

internal/yunikorn/sync.go
```diff
@@ -260,55 +208,57 @@ func (s *Service) upsertPartitionNodes(ctx context.Context, partitions []*dao.Pa
...
-
-	// Protect the applications and errors slices with a mutex so multiple goroutines can safely append queues
-	mutex := sync.Mutex{}
-
-	var errs []error
-	var apps []*dao.ApplicationDAOInfo
+	applications, err := s.client.GetApplications(ctx, "", "")
+	if err != nil {
+		return fmt.Errorf("could not get applications: %v", err)
+	}
```
sudiptob2, line 218: This will only give us the applications in the `default` partition. Should we loop over all the partitions and do this sync logic?

internal/database/repository/application.go
```diff
@@ -51,62 +48,240 @@ func applyApplicationFilters(builder *sql.Builder, filters ApplicationFilters) {
...
+	groups,
+	rejected_message,
+	state_log,
+	place_holder_data,
+	has_reserved,
+	reservations,
+	max_request_priority
+FROM
+	applications
+ORDER BY app_id, id DESC
```
sudiptob2, line 188: Just a quick question: Why do we need to order by `app_id`? Shouldn't the `id` alone be sufficient to keep the active ones at the top?

internal/database/repository/application.go
```diff
@@ -51,62 +48,240 @@ func applyApplicationFilters(builder *sql.Builder, filters ApplicationFilters) {
...
+LIMIT 1`
```
sudiptob2, line 189: I think we should not apply limit here.

internal/yunikorn/event_handler.go
```diff
@@ -37,234 +34,100 @@ func (s *Service) handleEvent(ctx context.Context, ev *si.EventRecord) error {
...
 	}
-}
 
-func (s *Service) handleAppRemoveEvent(ctx context.Context, ev *si.EventRecord) {
-	logger := log.FromContext(ctx)
+	app, err := s.repo.GetLatestApplicationByApplicationID(ctx, daoApp.ApplicationID)
+	if err != nil {
+		logger.Errorf("could not get application by application id: %v", err)
+		return
+	}
```
sudiptob2, line 72: There could be the same `app_id` present in multiple partitions. In such cases, `s.repo.GetLatestApplicationByApplicationID` might return a different application than the one we intend to update. Did we decide on having a convention of maintaining a globally unique `app_id`?

internal/model/application.go
```diff
@@ -0,0 +1,69 @@
...
+	app.PlaceholderData = appInfo.PlaceholderData
+
+	lookup := make(map[string]struct{})
+	if len(appInfo.Allocations) > 0 {
+		for _, alloc := range app.Allocations {
+			lookup[alloc.AllocationKey] = struct{}{}
+		}
+		for _, alloc := range appInfo.Allocations {
+			if _, ok := lookup[alloc.AllocationKey]; !ok {
+				app.Allocations = append(appInfo.Allocations, alloc)
```
sudiptob2, line 35: Should it be `app.Allocations`?
```suggestion
				app.Allocations = append(app.Allocations, alloc)
```

internal/model/application.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+		clear(lookup)
+	}
+	if len(appInfo.Requests) > 0 {
+		for _, ask := range app.Requests {
+			lookup[ask.AllocationKey] = struct{}{}
+		}
+		for _, ask := range app.Requests {
+			if _, ok := lookup[ask.AllocationKey]; !ok {
+				app.Requests = append(appInfo.Requests, ask)
```
sudiptob2, line 46:
```suggestion
				app.Requests = append(app.Requests, ask)
```

internal/database/repository/application.go
nikola-jokic: Good question and no, it wouldn't be sufficient.
Imagine this scenario:
I create app with `id=1`. The app finishes and it is removed. Then, someone creates the app with `id=1`. Since there is no app in the scheduler with `id=1`, it is fine from the yunikorn perspective. But this is completely new app. We would get create event to insert the new app.
In yunikorn, app_id is unique, since the app is going to be removed after it completes. But for yhs, app ID doesn't have to be unique. If it were the case, then we wouldn't need our own ID field.
What I'm trying to do is get the latest version of the application with this app_id.

internal/database/repository/application.go
nikola-jokic: Good catch!

internal/yunikorn/sync.go
nikola-jokic: All applications are applied to the default partition so it is okay to fetch it like this

internal/yunikorn/stream.go
nikola-jokic: I don't think we should do anything about it unless we have evidence it is a bottleneck. We are not doing something complicated, so I would avoid implementing parallel processing for now

internal/yunikorn/event_handler.go
nikola-jokic: All applications belong to the default partition

internal/database/repository/partitions.go
```diff
@@ -43,52 +40,62 @@ func applyPartitionFilters(builder *sql.Builder, filters PartitionFilters) {
...
-				"total_nodes":                p.TotalNodes,
-				"applications":               p.Applications,
-				"total_containers":           p.TotalContainers,
-				"state":                      p.State,
-				"last_state_transition_time": p.LastStateTransitionTime,
-			})
-		if err != nil {
-			return fmt.Errorf("could not insert/update partition into DB: %v", err)
-		}
+func (r *PostgresRepository) CreatePartition(ctx context.Context, partition *model.Partition) error {
```
sudiptob2, line 43: Should we rename to `InsertPartition`? Everywhere else it is InsertXX. Not a blocker though

nikola-jokic: Yep we should

internal/database/repository/queue.go
```diff
@@ -3,176 +3,65 @@ package repository
...
-		}
-	}
-	return nil
-}
-
-func (s *PostgresRepository) AddQueues(ctx context.Context, parentId *string, queues []*dao.PartitionQueueDAOInfo) error {
+func (s *PostgresRepository) InsertQueue(ctx context.Context, q *model.Queue) error {
 	insertSQL := `INSERT INTO queues (
-		parent_id, queue_name, status, partition, pending_resource, max_resource,
+		id, created_at_nano, queue_name, parent_id, status, partition, pending_resource, max_resource,
```
sudiptob2, line 14: I think this `parent_id` is not inserted properly. `buildPartitionQueueTrees` method in the webservice depends on this. Now `buildPartitionQueueTrees` might give a flat list rather than a tree.

nikola-jokic: We need @richscott 's changes to set the id so we can reliably set these parameters

internal/yunikorn/event_handler.go
```diff
@@ -83,32 +83,48 @@ func (s *Service) handleAppEvent(ctx context.Context, ev *si.EventRecord) {
...
+				ID:            ulid.Make().String(),
+				CreatedAtNano: ev.TimestampNano,
+			},
+			PartitionQueueDAOInfo: daoQueue,
+		}
+
+		if err := s.repo.InsertQueue(ctx, queue); err != nil {
+			logger.Errorf("could not insert queue: %v", err)
+			return
+		}
```
sudiptob2, line 110: It appears that the following insert will always fail due to a `NOT NULL` constraint on the `partition` field in the database (`partition TEXT NOT NULL CHECK (partition <> '')`).

YuniKorn does not provide a `partition` value for child queues, which results in an empty string (`""`) for the `partition` field when we receive events `state`. As a result, inserting these records will cause a DB error.

sudiptob2: For the initial sync it works because we update the `Partition` with the parent partition for all children in `flattenQueues()` method

internal/database/repository/queue.go
richscott: I am literally adding that to the `Queue` type and funcs right now, then just a few other types, in yk core. It will arrive very soon.

sudiptob2 (approved):
LGTM 🚀

PR #275, opened 2024-10-15
Add event `Node` event handler

### LINKED ISSUE #276: Add event handler for `Node` Events.
`Node` change events are now available in the event stream of the G-Research fork of yunikorn. Implement event handler for Node Events.

sudiptob2:
Closes #276 

Based on #271


internal/database/repository/node.go
```diff
@@ -57,48 +59,168 @@ func applyNodeFilters(builder *sql.Builder, filters NodeFilters) {
...
+func (s *PostgresRepository) DeleteNodesNotInIDs(ctx context.Context, ids []string, deletedAtNano int64) error {
+	const q = `
+UPDATE nodes 
+SET deleted_at_nano = @deleted_at_nano
+WHERE deleted_at_nano IS NULL AND NOT (id = ANY(@ids))`
+	_, err := s.dbpool.Exec(
+		ctx,
+		q,
+		pgx.NamedArgs{
+			"ids":             ids,
```
richscott, line 217: Can you swap the order of these two lines, so they match the order of the params in the SQL? It would be a little bit easier to follow.

richscott (commented):
Just one minor style suggestion for now, otherwise it looks great.

internal/database/repository/node.go
sudiptob2: updated

PR #277, opened 2024-10-15
Print queue state and update dependencies

nikola-jokic:
This PR updates the dependency and prints the state attribute from the event stream

Branch 1 of breaking down PR #271


dcoric (approved):
LGTM

PR #281, opened 2024-10-16
Application sync

### LINKED ISSUE #272: Add event-handler logic for EventRecord_APP  type
The event stream now propagates data through the state. Implement the handler that will synchronize application states.

nikola-jokic:
Fixes #272

### DIFF
--- a/internal/database/repository/node.go
+++ b/internal/database/repository/node.go
@@ -103,7 +103,6 @@ func (s *PostgresRepository) InsertNodeUtilizations(
 	ctx context.Context,
 	nus []*dao.PartitionNodesUtilDAOInfo,
 ) error {
-
 	insertSQL := `INSERT INTO partition_nodes_util (id, cluster_id, partition, nodes_util_list)
 		VALUES (@id, @cluster_id, @partition, @nodes_util_list)`
 
@@ -138,7 +137,6 @@ func (s *PostgresRepository) GetNodeUtilizations(
 	query := queryBuilder.Query()
 	args := queryBuilder.Args()
 	rows, err := s.dbpool.Query(ctx, query, args...)
-
 	if err != nil {
 		return nil, fmt.Errorf("could not get node utilizations from DB: %v", err)
 	}
--- a/internal/model/application.go
+++ b/internal/model/application.go
@@ -5,7 +5,7 @@ import (
 )
 
 type Application struct {
-	ModelMetadata          `json:",inline"`
+	Metadata               `json:",inline"`
 	dao.ApplicationDAOInfo `json:",inline"`
 }
 
--- a/internal/model/application.go
+++ b/internal/model/application.go
@@ -1,0 +1,69 @@
+package model
+
+import (
+	"github.com/G-Research/yunikorn-core/pkg/webservice/dao"
+)
+
+type Application struct {
+	ModelMetadata          `json:",inline"`
+	dao.ApplicationDAOInfo `json:",inline"`
+}
+
+func (app *Application) MergeFrom(appInfo *dao.ApplicationDAOInfo) {
+	app.Partition = appInfo.Partition
+	app.QueueName = appInfo.QueueName
+	app.SubmissionTime = appInfo.SubmissionTime
+	app.FinishedTime = appInfo.FinishedTime
+	app.State = appInfo.State
+	app.User = appInfo.User
+	app.Groups = appInfo.Groups
+	app.RejectedMessage = appInfo.RejectedMessage
+	app.StateLog = appInfo.StateLog
+	app.HasReserved = appInfo.HasReserved
+	app.Reservations = appInfo.Reservations
+	app.MaxRequestPriority = appInfo.MaxRequestPriority
+	app.StartTime = appInfo.StartTime
+	app.PlaceholderData = appInfo.PlaceholderData
+
+	lookup := make(map[string]struct{})
+	if len(appInfo.Allocations) > 0 {
+		for _, alloc := range app.Allocations {
+			lookup[alloc.AllocationKey] = struct{}{}
+		}
+		for _, alloc := range appInfo.Allocations {
+			if _, ok := lookup[alloc.AllocationKey]; !ok {
+				app.Allocations = append(app.Allocations, alloc)
+			}
+		}
+		clear(lookup)
+	}
+	if len(appInfo.Requests) > 0 {
+		for _, ask := range app.Requests {
+			lookup[ask.AllocationKey] = struct{}{}
+		}
+		for _, ask := range app.Requests {
+			if _, ok := lookup[ask.AllocationKey]; !ok {
+				app.Requests = append(app.Requests, ask)
+			}
+		}
+		clear(lookup)
+	}
+	if appInfo.UsedResource != nil {
+		app.UsedResource = appInfo.UsedResource
+	}
+	if appInfo.MaxUsedResource != nil {
+		app.MaxUsedResource = appInfo.MaxUsedResource
+	}
+	if appInfo.PendingResource != nil {
+		app.PendingResource = appInfo.PendingResource
+	}
+	if appInfo.ResourceUsage != nil {
+		app.ResourceUsage = appInfo.ResourceUsage
+	}
+	if appInfo.PreemptedResource != nil {
+		app.PreemptedResource = appInfo.PreemptedResource
+	}
+	if appInfo.PlaceholderResource != nil {
+		app.PlaceholderResource = appInfo.PlaceholderResource
+	}
+}
--- a/internal/model/application.go
+++ b/internal/model/application.go
@@ -41,7 +41,7 @@ func (app *Application) MergeFrom(appInfo *dao.ApplicationDAOInfo) {
 		for _, ask := range app.Requests {
 			lookup[ask.AllocationKey] = struct{}{}
 		}
-		for _, ask := range app.Requests {
+		for _, ask := range appInfo.Requests {
 			if _, ok := lookup[ask.AllocationKey]; !ok {
 				app.Requests = append(app.Requests, ask)
 			}
--- a/internal/model/models.go
+++ b/internal/model/models.go
@@ -4,7 +4,7 @@ import (
 	"github.com/G-Research/yunikorn-core/pkg/webservice/dao"
 )
 
-type ModelMetadata struct {
+type Metadata struct {
 	ID        string `json:"id"`
 	CreatedAt int64  `json:"createdAt"`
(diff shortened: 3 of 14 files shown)


internal/model/application.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	if appInfo.ResourceUsage != nil {
+		app.ResourceUsage = appInfo.ResourceUsage
+	}
+	if appInfo.PreemptedResource != nil {
+		app.PreemptedResource = appInfo.PreemptedResource
+	}
+	if appInfo.PlaceholderResource != nil {
+		app.PlaceholderResource = appInfo.PlaceholderResource
+	}
```
sudiptob2, line 68: I was wondering do we need these nil checks?

internal/database/repository/node.go
```diff
@@ -138,7 +137,6 @@ func (s *PostgresRepository) GetNodeUtilizations(
 	query := queryBuilder.Query()
 	args := queryBuilder.Args()
 	rows, err := s.dbpool.Query(ctx, query, args...)
-
 	if err != nil {
 		return nil, fmt.Errorf("could not get node utilizations from DB: %v", err)
```
dejanzele, line 141: I'd suggest using `%w` so errors are wrapped and `errors.Is` works correctly, even though we may not have a concrete use case, this can often lead to silent bugs

internal/model/application.go
```diff
@@ -0,0 +1,69 @@
...
+	app.Groups = appInfo.Groups
+	app.RejectedMessage = appInfo.RejectedMessage
+	app.StateLog = appInfo.StateLog
+	app.HasReserved = appInfo.HasReserved
+	app.Reservations = appInfo.Reservations
+	app.MaxRequestPriority = appInfo.MaxRequestPriority
+	app.StartTime = appInfo.StartTime
+	app.PlaceholderData = appInfo.PlaceholderData
+
+	lookup := make(map[string]struct{})
```
dejanzele, line 28: I'd suggest `any` instead of `struct{}`

internal/model/application.go
```diff
@@ -0,0 +1,69 @@
...
+	if len(appInfo.Allocations) > 0 {
+		for _, alloc := range app.Allocations {
+			lookup[alloc.AllocationKey] = struct{}{}
```
dejanzele, line 31: Why can't this just be `nil`?

internal/model/application.go
```diff
@@ -0,0 +1,69 @@
...
+		for _, alloc := range appInfo.Allocations {
+			if _, ok := lookup[alloc.AllocationKey]; !ok {
+				app.Allocations = append(app.Allocations, alloc)
+			}
+		}
+		clear(lookup)
+	}
+	if len(appInfo.Requests) > 0 {
+		for _, ask := range app.Requests {
+			lookup[ask.AllocationKey] = struct{}{}
```
dejanzele, line 42: same as above

internal/model/application.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+		for _, ask := range app.Requests {
+			if _, ok := lookup[ask.AllocationKey]; !ok {
+				app.Requests = append(app.Requests, ask)
+			}
+		}
+		clear(lookup)
+	}
+	if appInfo.UsedResource != nil {
```
dejanzele, line 51: What is the difference than just doing:
```
app.UsedResource = appInfo.UsedResource
```

I think the if is redundant, if it is nil, it will also be nil in `app`.

internal/yunikorn/event_handler.go
```diff
@@ -33,206 +30,53 @@ func (s *Service) handleEvent(ctx context.Context, ev *si.EventRecord) error {
...
-		s.handleAppSetEvent(ctx, ev)
-	case si.EventRecord_REMOVE:
-		s.handleAppRemoveEvent(ctx, ev)
-	case si.EventRecord_NONE:
-	default:
-		// should be warning
-		logger.Warnf("unknown event EventChangeType for an Event of type APP: %v", ev.GetEventChangeType())
+	var daoApp dao.ApplicationDAOInfo
+	if err := json.Unmarshal([]byte(ev.GetState()), &daoApp); err != nil {
+		logger.Errorw("Failed to unmarshal application state from event", "error", err)
```
dejanzele, line 41: rest od the code uses lowercase for `Errorw`

internal/yunikorn/sync.go
```diff
@@ -260,55 +251,54 @@ func (s *Service) upsertPartitionNodes(ctx context.Context, partitions []*dao.Pa
...
-	wg.Wait()
+	now := time.Now().UnixNano()
+	for _, a := range applications {
+		current, ok := lookup[a.ApplicationID]
+		delete(lookup, a.ApplicationID)
+		if !ok || current.DeletedAt != nil { // either not exists or deleted
+			application := &model.Application{
+				ModelMetadata: model.ModelMetadata{
+					ID:        ulid.Make().String(),
+					CreatedAt: now,
```
dejanzele, line 281: Should all of them have the same `createdAt`?

internal/yunikorn/sync.go
```diff
@@ -260,55 +251,54 @@ func (s *Service) upsertPartitionNodes(ctx context.Context, partitions []*dao.Pa
...
-	// wait for all queues to be processed
-	wg.Wait()
+	now := time.Now().UnixNano()
+	for _, a := range applications {
+		current, ok := lookup[a.ApplicationID]
+		delete(lookup, a.ApplicationID)
+		if !ok || current.DeletedAt != nil { // either not exists or deleted
+			application := &model.Application{
+				ModelMetadata: model.ModelMetadata{
+					ID:        ulid.Make().String(),
```
dejanzele, line 280: I'd suggest having a `NewModelMetadata` which has this logic for creating IDs instead of manually instantiating it, but that is up to you

internal/model/models.go
```diff
@@ -1,15 +1,13 @@
...
-	"time"
-
 	"github.com/G-Research/yunikorn-core/pkg/webservice/dao"
 )
 
-type ApplicationDAOInfo struct {
-	CreatedAt              time.Time `json:"createdAt"`
-	QueueID                string    `json:"queueId"`
-	dao.ApplicationDAOInfo `json:",inline"`
+type ModelMetadata struct {
```
dejanzele, line 7: Why not name it just `Metadata`? It is clear it is a model from the package name

internal/model/application.go
nikola-jokic: struct has 0 size while any has 2 fields in a struct. Since I don't really care about the value, and this is just a set, it is always more readable to me when I see `map[string]struct{}` to know what is it used for.

internal/model/application.go
nikola-jokic: Doesn't have to be
App accumulates stuff during many events, so if during deletion, used resources is set to nil, I want to preserve the old data. App is persisted in the database, and later merged on event.

internal/database/repository/node.go
nikola-jokic: The intention is to modify the app, these are just formatting changes... I can apply the suggestion once we sync the nodes if you'd like. :relaxed:

internal/model/models.go
```diff
@@ -1,15 +1,13 @@
...
+	ID        string `json:"id"`
+	CreatedAt int64  `json:"createdAt"`
```
dejanzele, line 9: If it isn't a `time.Time`, can we name it `CreatedAtNano` or `CreatedAtMilli` so the precision is not ambiguous.

internal/model/models.go
nikola-jokic: I forgot to do that... Good catch!

internal/model/models.go
```diff
@@ -1,15 +1,13 @@
...
+	DeletedAt *int64 `json:"deletedAt,omitempty"`
```
dejanzele, line 10: same as above

internal/yunikorn/sync.go
nikola-jokic: Yes, because that is the time when we synced stuff. The application data from the scheduler is within the dao object, while the best approximation for when we are creating it is during the sync time. Since we cannot get the event timestamp at this time, I wanted to use sync time to determine when we fetched the data from the yunikorn API.

internal/yunikorn/sync.go
```diff
@@ -260,55 +251,54 @@ func (s *Service) upsertPartitionNodes(ctx context.Context, partitions []*dao.Pa
 	return nil
 }
 
-// upsertApplications fetches applications for each queue and upserts them into the database
-func (s *Service) upsertApplications(ctx context.Context, queues []*dao.PartitionQueueDAOInfo) error {
+// syncApplications fetches applications for each queue and upserts them into the database
+func (s *Service) syncApplications(ctx context.Context) error {
```
sudiptob2, line 255: Might be a good idea to a some tests for this method.

internal/model/application.go
nikola-jokic: Here is the answer: https://github.com/G-Research/yunikorn-history-server/pull/281#discussion_r1803260935 :relaxed:

internal/model/application.go
```diff
@@ -0,0 +1,69 @@
...
+	app.QueueName = appInfo.QueueName
+	app.SubmissionTime = appInfo.SubmissionTime
+	app.FinishedTime = appInfo.FinishedTime
+	app.State = appInfo.State
+	app.User = appInfo.User
+	app.Groups = appInfo.Groups
+	app.RejectedMessage = appInfo.RejectedMessage
+	app.StateLog = appInfo.StateLog
+	app.HasReserved = appInfo.HasReserved
+	app.Reservations = appInfo.Reservations
```
sudiptob2, line 23: Do we need logic similar to Allocations for merging Reservations?

internal/model/application.go
```diff
@@ -0,0 +1,69 @@
...
+	app.MaxRequestPriority = appInfo.MaxRequestPriority
+	app.StartTime = appInfo.StartTime
+	app.PlaceholderData = appInfo.PlaceholderData
```
sudiptob2, line 26: Do we need logic similar to Allocations for merging PlaceholderData?

sudiptob2 (approved):
lgtm

internal/model/models.go
richscott: +1 on the renaming suggestion.

richscott (commented):
Only thing I'd suggest is that timestamp renaming that Dejan suggested, but I won't stand in the way if you'd rather not do that.

dcoric (approved):
LGTM

PR #296, opened 2024-10-25
Fix application checks with the new id field

nikola-jokic:
This change marks the old applications as deleted and avoids fetching them from the db


sudiptob2 (approved):
LGTM

PR #314, opened 2024-11-01
Extend applications model to include partition ID

### LINKED ISSUE #313: Add partition ID to the application now that application model is extended
The schema for `application` is missing the partition_id field added after the application event stream sync.
We should insert the partition ID along with other fields.

nikola-jokic:
Fixes #313


sudiptob2 (approved):
LGTM

PR #317, opened 2024-11-01
Switch routes to query by id instead of partition_name or queue_name

### LINKED ISSUE #316: Switch routes from object_name to object_id
Since object names are not unique, we should query objects by their IDs.

Depends on https://github.com/G-Research/yunikorn-core/pull/5

nikola-jokic:
Fixes #316


richscott (approved):
Looks good.

internal/database/repository/application.go
```diff
@@ -327,11 +333,11 @@ func (s *PostgresRepository) GetAllApplications(ctx context.Context, filters App
...
 
 //nolint:all
-func (s *PostgresRepository) GetAppsPerPartitionPerQueue(ctx context.Context, partition, queue string, filters ApplicationFilters) ([]*model.Application, error) {
+func (s *PostgresRepository) GetAppsPerPartitionPerQueue(ctx context.Context, partitionID, queueID string, filters ApplicationFilters) ([]*model.Application, error) {
 	queryBuilder := sql.NewBuilder().
 		SelectAll("applications", "").
-		Conditionp("queue_name", "=", queue).
-		Conditionp("partition", "=", partition).
+		Conditionp("queue_id", "=", queueID).
+		Conditionp("partition_id", "=", partitionID).
```
sudiptob2, line 340: Just a ques, If we filter based on `queue_id` do we need `partition_id` ?

nikola-jokic: Good question
My opinion is, since we have a route being /partition/partition_id/queue/queue_id, it doesn't hurt
We can change the route and remove the partition ID, but I wouldn't want to have route with partition_id and never use that parameter

sudiptob2: Hmm..🤔 I was wondering if it is worth mirroring the `yunikorn` route parameters then, if we are going to use IDs.

PR #322, opened 2024-11-04
Feat/288/use fullstate dump

### LINKED ISSUE #288: Source state before the event handler takes place

sudiptob2:
closes #288


internal/yunikorn/rest.go
```diff
@@ -29,6 +30,15 @@ type RESTClient struct {
 	port     int
 }
 
+type AggregatedStateInfo struct {
```
nikola-jokic, line 33: Why not use yunikorn-core `webservice.AggregatedStateInfo`? We depend already on yunikorn-core, so why not simply re-use their structure that is public?

internal/yunikorn/sync.go
```diff
@@ -9,17 +9,18 @@ import (
...
-	partitions, err := s.client.GetPartitions(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("could not get partitions: %v", err)
+func (s *Service) syncPartitions(ctx context.Context, partitions []*dao.PartitionInfo) error {
+	if partitions == nil {
+		var err error
+		partitions, err = s.client.GetPartitions(ctx)
+		if err != nil {
+			return fmt.Errorf("could not get partitions: %w", err)
+		}
```
nikola-jokic, line 23: Why would we fetch partitions if they are not passed in?
I'd just do `if len(partitions) == 0 { return nil }`

internal/yunikorn/sync.go
```diff
@@ -224,16 +187,10 @@ func (s *Service) syncApplications(ctx context.Context) error {
...
-
-	appsHistory, err := s.client.GetAppsHistory(ctx)
-	if err != nil {
-		return fmt.Errorf("could not get apps history: %w", err)
-	}
-	containersHistory, err := s.client.GetContainersHistory(ctx)
-	if err != nil {
-		return fmt.Errorf("could not get containers history: %w", err)
-	}
+func (s *Service) syncHistory(
```
nikola-jokic, line 190: We should probably split this into syncAppHistory and syncContainerHistory. is there a good reason why this method is handling multiple things?

nikola-jokic (changes requested):
Wrote few questions :relaxed: Overall looks good

internal/yunikorn/sync.go
sudiptob2: `syncPartitions()` is also used in the accumulator, because we do not have events for partition in the event stream. 
ref: https://github.com/G-Research/unicorn-history-server/blob/545795dfc935636c1c0b9a4a9a80e7105f7e269e/internal/yunikorn/service.go#L43-L50

internal/yunikorn/sync.go
sudiptob2: yeah, I was thinking about it, Actually was designed this way from the beginning so I just carried it forward.

internal/yunikorn/sync.go
nikola-jokic: Oh that part should be removed completely.
When we sync queues, we should check if partition exists and sync partitions if not

richscott: Agree - syncing partitions on every EventRecord received would be expensive.

sudiptob2: If we only sync when the partition does not exist, we cannot handle 'partition update' scenarios. I believe we should sync the partition for all queue events, regardless of whether the partition already exists.

internal/yunikorn/accumulator.go
```diff
@@ -47,7 +47,7 @@ func (a *accumulator) add(event *si.EventRecord) {
 }
 
 // run is responsible for running the accumulator.
-func (a *accumulator) run(ctx context.Context) error {
+func (a *accumulator) run(ctx context.Context) {
```
nikola-jokic, line 50: We should remove this completely. We don't need this anymore since if queue doesn't exist, we should sync partitions

internal/yunikorn/event_handler.go
```diff
@@ -89,12 +89,21 @@ func (s *Service) handleQueueEvent(ctx context.Context, ev *si.EventRecord) {
 		return
 	}
 
+	// Sync partitions before syncing queues
+	partitions, err := s.client.GetPartitions(ctx)
```
nikola-jokic, line 93: We should check if partition exists. If not, only then we should sync

nikola-jokic (changes requested):
Few more changes before we are done with this change :relaxed:

internal/yunikorn/event_handler.go
sudiptob2: What if the partition exists in our database, but some of its data has changed on the YuniKorn side? If we only sync when the partition doesn’t exist in our database,  we will end up with the initial state of the partition always.

richscott (dismissed):

richscott (approved):
lgtm

nikola-jokic (approved):

PR #329, opened 2024-11-07
Fix UHS performance-tests target and JS test code

richscott:
- Fix UHS container instantiation via helm
- Change 'disown' shell usage for detaching UHS process to use 'nohup' wrapper prefix instead - 'disown' is available in full Bash, but not minimal /bin/sh versions.
- Fix JS test code to use UHS-style /api/v1/... API prefix, and use correct status_text for parsing response.
- Use recent yunikorn-core (GR fork) image version - this is fixed for now, but we will change it to `latest` once that tag is published for yk-core.

Fixes https://github.com/G-Research/unicorn-history-server/issues/309


nikola-jokic (approved):
lgtm

nikola-jokic (changes requested):
Wait, integration tests are failing, can you please fix that

Makefile
```diff
@@ -208,7 +208,7 @@ define start-cluster
 	@echo "**********************************"
 	@echo "Creating cluster"
 	@echo "**********************************"
-	@CLUSTER_NAME=uhs-test $(MAKE) create-cluster
+	$(MAKE) create-cluster
```
sudiptob2, line 211: Any specific reason for removing `@CLUSTER_NAME=uhs-test`? Creating a separate cluster for test seemed to be a good idea.

richscott: Okay - reverted back in commit 70d0db9

richscott:
> Wait, integration tests are failing, can you please fix that

All tests finally work now! Please re-review.

Makefile
```diff
@@ -97,6 +97,10 @@ KIND_VERSION ?= latest
 MINIKUBE ?= $(LOCALBIN_TOOLING)/minikube
 MINIKUBE_VERSION ?= latest
 
+# The release version of Yunikorn images
+# used in integration and performance tests.
+YK_VERSION=fab384e
```
sudiptob2, line 102: I just have a small question.
So, we have to change this version every time we release a YK image.
Why don't we release with the `latest` tag?

sudiptob2 (approved):

nikola-jokic (approved):

Makefile
dave-gantenbein: I always wonder that!

richscott: We will change this to `latest` after we unleash our first `v1.6.0-1-gr0` (or whatever) image release, and @pavlovic-ivan can also make a `latest` tag alias to that.

PR #332, opened 2024-11-12
implement testsuite and test-containers for repository integration test

### LINKED ISSUE #82: Add setup/teardown logic for running multiple integration tests in same test instance
As we add more integration tests to YHS, it may cause the overall test run to become quite a bit slower if each test function does its own setup of the Yunikorn server, Postgres db, YHS, etc, and then shutdown/terminate each one when the test func is finished. We should consider creating some helper logic that does a single, one-time setup of the server(s), then have all the integration tests run, then a single tear-down operation. (To ensure a clean state for test funcs, we may not want to do this for all services, or maybe have helper(s) to clear the YHS database between test funcs.)

We can use the Go standard library's `TestMain` func support (see https://pkg.go.dev/testing) to simulate one-time setup/teardown, or we could use the `TestSuite` facility in the `testify` library (we already use `testify/assert`) to set up structured sets of tests and easily declare helper funcs to run before/after both test-sets and individual test funcs - see https://pkg.go.dev/github.com/stretchr/testify/suite .

sudiptob2:
closes #82


internal/database/repository/node_int_test.go
```diff
@@ -8,29 +8,52 @@ import (
...
-	if testing.Short() {
-		t.Skip("skipping integration test in short mode")
-	}
+type NodeTestSuite struct {
+	suite.Suite
+	tp   *database.TestPostgresContainer
+	repo *PostgresRepository
+}
 
+func (ns *NodeTestSuite) SetupSuite() {
```
nikola-jokic, line 25: Do we want to run many test suites? Isn't it better to have a single db instance and maybe, if we really need to do it, create different databases?

sudiptob2: Thanks. I thought creating multiple db containers would help to run the tests in parallel. But the idea of creating multiple db in a single container sounds good.

sudiptob2: updated with separate DB instead of separate container

test/database/postgres.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+type InstanceConfig struct {
+	User     string
+	Password string
+	DBName   string
+	Host     string
+	Port     string
```
richscott, line 23: What do you think about making this an `int` instead of a string, for a little more type-safety?

richscott (commented):
It looks good - I just had one type suggestion.

test/database/postgres.go
sudiptob2: updated

richscott (approved):

richscott (approved):

PR #343, opened 2024-11-19
Refactor sync integration test with test containers.

### LINKED ISSUE #339: Refactor legacy integration test using Testsuite and Testcontainers
After the implementation of testsuite in PR https://github.com/G-Research/unicorn-history-server/pull/332 we now need to refactor remain integration test using the test suite.

sudiptob2:
closes #339


internal/health/health_topsuite_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"context"
+	"testing"
+
+	"github.com/G-Research/unicorn-history-server/test/database"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+)
+
+type HealthToptSuite struct {
```
richscott, line 13: I don't understand this naming - specifically, the "Topt" part of "HealthToptSuite" - what does that signify?

internal/yunikorn/sync_nodes_int_test.go
```diff
@@ -0,0 +1,121 @@
...
+		expectedNodes []*model.Node
+		wantErr       bool
+	}{
+		{
+			name: "Sync nodes with no existing nodes",
+			stateNodes: []*dao.NodesDAOInfo{
+				{
+					PartitionName: "default",
+					Nodes: []*dao.NodeDAOInfo{
+						{ID: "1", NodeID: "node-1", Partition: "default", HostName: "host-1"},
```
richscott, line 50: Should the `Partition`s here be `PartionID` instead?

internal/yunikorn/sync_nodes_int_test.go
```diff
@@ -0,0 +1,121 @@
...
+			},
+			wantErr: false,
+		},
+		{
+			name: "Sync nodes with existing nodes in DB",
+			stateNodes: []*dao.NodesDAOInfo{
+				{
+					PartitionName: "default",
+					Nodes: []*dao.NodeDAOInfo{
+						{ID: "2", NodeID: "node-2", Partition: "default", HostName: "host-2-updated"},
```
richscott, line 68: Same here - should `PartitionID` be used instead?

richscott (commented):
Just a few questions about use of `PartitionID` instead of `Partition` - there might be errors if this branch has not merged from `main` recently.

internal/health/health_topsuite_test.go
sudiptob2: That's a good catch, Wanted to write `HealthTopSuite` but misspelled! Still not satisfied with the name though. Let me know if you can think of a better name 🤞

richscott: How about just `HealthSuite`?  Its location in the hierarchy hints that it will be running sub-suites, etc.

sudiptob2: Yeah make sense, updated

internal/yunikorn/sync_nodes_int_test.go
sudiptob2: Updated after merge

internal/yunikorn/sync_nodes_int_test.go
sudiptob2: Updated after merge

richscott (approved):

PR #355, opened 2024-11-26
Rewrite gomigrate test using testcontainer

### LINKED ISSUE #347: Refactor `GoMigrate_Integration` test using Testcontainer

sudiptob2:
closes #347


sudiptob2:
> Why do we need to test migration when migration is a requirement for all other tests?

Hmm.. Interesting question. It tests `migration.Up()` and `migration.Down()` commands that we have. For that reason I think we should keep this test.

nikola-jokic (approved):

PR #359, opened 2024-12-02
Clean up and make cluster Make targets consistent.

richscott:
Change Makefile targets for running the Yunikorn services to make testing a separate UHS instance a bit easier. Developers can now use `make create-cluster` and `make delete-cluster`, to be consistent with Kind's `create-cluster` and `delete-cluster` commands.

Fixes https://github.com/G-Research/unicorn-history-server/issues/356


sudiptob2 (approved):
LGTM

PR #366, opened 2024-12-11
Improve error response when resource is unavailable

### LINKED ISSUE #357: UHS get-applications endpoint issues
If the K8S cluster has just started up, and the Yunikorn scheduler is still finishing its setup, starting the UHS service immediately may result in an initial-sync error ("partition not found") - this message should not be printed, as it simply means that the partition(s) found from YK-core do not yet exist at all in the (new) UHS database.

The UHS logs, when starting it with `make run`, show this near the beginning:
```
go run cmd/unicorn-history-server/main.go --config config/unicorn-history-server/local.yml
2024-12-04T10:57:58.895-0700    INFO    starting unicorn history server {"version": "0.0.0-dev", "buildTime": "2024-12-04T10:57:58-07:00", "commit": "unknown"}
2024-12-04T10:57:58.900-0700    INFO    starting webservice on :8989    {"component": "webservice"}
Error getting partition: could not get partition from DB: no rows in result set
2024-12-04T10:57:58.955-0700    INFO    starting yunikorn event stream client   {"component": "yunikorn_event_collector"}
```

The second issue that should be improved/fixed is the applications endpoint - if the client specifies a partition-id or queue-id that does not exist, then the endpoint should return an error message (possibly with a 404 status) that the queue-id is invalid and/or the partition-id is invalid.

The third issue is that if no applications are found (whether or not the queue/partitions exist) is that currently it returns just a bare `null` in the response body - this should instead be `[]` so that it is valid JSON, and is parseable by the client.

sudiptob2:
Closes #357  

- [x] Removed unnecessary `print` statement for errors.  
- [x] Updated application endpoint to include a standard 404 response when a resource is not found.  
- [x] Updated other endpoints in the web service to follow a similar approach by adding a standard 404 response.

### DIFF
--- a/internal/yunikorn/sync.go
+++ b/internal/yunikorn/sync.go
@@ -28,7 +28,6 @@ func (s *Service) syncPartitions(ctx context.Context, partitions []*dao.Partitio
 	for _, p := range partitions {
 		current, err := s.repo.GetPartitionByID(ctx, p.ID)
 		if err != nil {
-			fmt.Printf("Error getting partition: %v\n", err)
 			partition := &model.Partition{
 				Metadata: model.Metadata{
 					CreatedAtNano: now,
--- a/internal/webservice/response.go
+++ b/internal/webservice/response.go
@@ -43,3 +43,17 @@ func badRequestResponse(req *restful.Request, resp *restful.Response, err error)
 		log.Logger.Errorf("could not write error response: %v", err)
 	}
 }
+
+func notFoundResponse(req *restful.Request, resp *restful.Response, err error) {
+	log.Logger.Errorf("error processing request for %s: %v", req.Request.URL.Path, err)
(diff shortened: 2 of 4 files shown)


richscott (approved):
This is a great improvement!

internal/yunikorn/sync.go
```diff
@@ -28,7 +28,6 @@ func (s *Service) syncPartitions(ctx context.Context, partitions []*dao.Partitio
 	for _, p := range partitions {
 		current, err := s.repo.GetPartitionByID(ctx, p.ID)
 		if err != nil {
-			fmt.Printf("Error getting partition: %v\n", err)
```
richscott, line 31: Wow, I thought we removed this already. Well, it's good to get it out now.

richscott (approved):

PR #395, opened 2025-01-15
Fix: YK Web build

dcoric:
Migrated Yunikorn Web build from package.json scripts to the make file.
Adds @types/node for setting env variables on build time.


sudiptob2 (commented):
Do we need to update the `build/unicorn-history-server/Dockerfile` ?

sudiptob2 (approved):

PR #404, opened 2025-01-22
Denis coric/prod env

dcoric:
This pull request is altering how UHS & YK Web use environment configuration.

- If production is active on build-time, both web apps will use information from the current window (path and port) to target the UHS API directly
- Override is still possible for development purposes (production flag set to false)

This change aims to remove hardcoded dependencies and paths on UHS/YK Web and allow deployment flexibility.


sudiptob2 (dismissed):
LGTM

sudiptob2 (approved):

